PR #12, opened 2024-05-23
Update pages.yml

nhayes-rich:
* Restrict doc remaking to relevant file changes
* Check docs make successfully on PR

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -4,6 +4,27 @@ on:
   # Runs on pushes targeting the default branch
   push:
     branches: ["main"]
+    paths:
+      - .github/workflows/pages.yml
+      - changelog/*
+      - docs/*
+      - example-notebooks/*
+      - src/seismometer/*
+      - Makefile
+      - pyproject.toml
+      - setup.cfg
+  
+  # Runs on PR but doesn't deploy
+  pull_request: 
+    paths:
+      - .github/workflows/pages.yml
+      - changelog/*
+      - docs/*
+      - example-notebooks/*
+      - src/seismometer/*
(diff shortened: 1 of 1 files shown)


.github/workflows/doc.yml
```diff
@@ -0,0 +1,68 @@
+name: Deploy Sphinx Doc to Pages
```
diehlbw: Do we need a second workflow for this? 
Conceptually it feels like build and deploy are related steps, and "proximty-wise" there is less magic if they are in the file as well

nhayes-rich: Think we might need to split them out if we want docs to be able to make on PR. pages.yml won't seem to run on a PR even if it's not going to deploy.

.github/workflows/pages.yml
```diff
@@ -4,7 +4,27 @@ on:
...
+      - example-notebooks/*
+      - src/seismometer/*
+      - Makefile
+      - pyproject.toml
+      - setup.cfg
+  
+  pull_request:
+    paths:
+      - .github/workflows/pages.yml
+      - changelog/*
```
diehlbw, line 20: Nit: I think we can drop the changelog reference; it's currently not in scope of the doc build (to my knowledge) so won't represent a pages change.

The release PR will maybe have some edits to make on the constructed release_notes, but I think that's ok. We can revisit if it becomes burdensome.

diehlbw (changes requested):

.github/workflows/pages.yml
nhayes-rich: finc

diehlbw (approved):

PR #15, opened 2024-05-24
CI Test suite for multiple enviroments

diehlbw:
# Overview
Updates CI for test coverage

## Description of changes
Coverage should be done on base environment (py3.10 + linux) since there is no dependent branching  
Test suite should run on:  
- linux, macos, windows  
- python 3.10 and 3.11

3.12 has known install issue to resolve before adding - marked as experimental allowing failure  

Using a runner matrix allowed for testing across environments and python versions **without** actually using nox.  While this isn't easily reproducible on a single machine, it seems reasonable to forgo defining the nox sessions until we need more complexity. Testing multiple environments in a single python version might be reason to start using sessions (such as ensuring compatibility with multiple versions of pandas)

## Author Checklist
- [X] Linting passes; run early with [pre-commit hook](https://pre-commit.com/#install).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 permissions: "read-all"
 
 jobs:
-  run-linting:
+  linting:
     name: linting via pre-commit 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


diehlbw:
> Are we planning on including the failing 3.12 pipeline still?
Removed, seems cleaner avoids the confusion of failing tasks

PR #23, opened 2024-06-06
Make label checks more dynamic

### LINKED ISSUE #22: Changelog workflow doesn't skip "late" label
## Problem Summary
Changes that do not warrant a changelog issue still fail if the label is added after the PR creation, until the next push

## Impact
The checks can fail in situations where they shouldn't

## Possible Solution
One option is to trigger on label changes, but there may be better methods.
Until fixed, whitespace commits will fully re-evaluate the checks

diehlbw:
# Overview

Closes #22 

## Description of changes
Use gh api to request labels dynamically instead of using the cached list that requires an event (commit) to renew. 
Split changelog workflow into two jobs; could not find a method for short-circuiting success within a job.  

Note: this method is intentionally requiring manual retry, and NOT triggering off of label changes

Suggested testing-  
As a step 0, I'm intentionally NOT setting labels prior to PR creation.  
Tests can add and remove labels on this PR, then manually rerun the specific workflow.  
Verify that the second job runs and is skipped depending on the skip-changelog label.  
Before merge, update labels and drop the second commit to verify the push event also behaves correctly

## Author Checklist

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -6,11 +6,28 @@ on:
 permissions: "read-all"
 
 jobs:
+  # Check for label in its own job to allow skipping multiple steps with 'success'
+  check-changelog-needed:
+    name: changelog needed
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Labels
+        id: check_labels
+        env:
+          OWNER: ${{ github.repository_owner }}
+          REPO_NAME: ${{ github.event.repository.name }}
+          PR_ID: ${{ github.event.pull_request.number }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          labels="$(gh api repos/$OWNER/$REPO_NAME/pulls/$PR_ID --jq '.labels.[].name')"
+          if [[ $labels != *"skip changelog"* ]]; then
+            echo "needs_entry=true" >> "$GITHUB_OUTPUT"
+          fi
+
   check-changelog-entry:
     name: changelog entry
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


diehlbw:
Initial run 'successfully' set needs_entry.  
Rerun after setting a different label, timed out on first attempt(?) then set it correctly.

.github/workflows/changelog.yml
```diff
@@ -6,11 +6,32 @@ on:
...
+            echo "needs_entry=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "needs_entry=false" >> "$GITHUB_OUTPUT"
+          fi
+    outputs:
+      needs_entry: ${{ steps.check-labels.outputs.needs_entry }}
+
   check-changelog-entry:
-    name: changelog entry
+    name: has fragment
```
firste, line 31: Can we make the name clearer for this? Like "has changelog fragment"?

.github/workflows/changelog.yml
```diff
@@ -6,11 +6,32 @@ on:
 permissions: "read-all"
```
firste, line 6: I think we only need "pull-requests: read" permission here.

firste (changes requested):

.github/workflows/changelog.yml
diehlbw: FINC

.github/workflows/changelog.yml
diehlbw: works for me

firste (approved):

PR #28, opened 2024-06-20
Rework plots return displayable objects, and allow caching as HTML/SVG

gbowlin:
# Overview
This PR is to update some of the plots to allow for easier caching when we switch between plots. 
It also makes plots a bit more uniform and enables theming by moving static titles from inside the plot to an HTML element. 

## Description of changes
- added `disk_cached_html_segments` for plots that are cached. 
- plots in `_api` now return displayable objects (HTML) rather than having plots get shown. 

## Author Checklist


src/seismometer/controls/decorators.py
```diff
@@ -2,13 +2,79 @@
...
+from inspect import signature
+from pathlib import Path
 from typing import Any, Callable
 
-from IPython.display import display
+from IPython.display import HTML
 from ipywidgets import Widget
 
+logger = logging.getLogger("seismometer")
+SEISMOMETER_CACHE_DIR = Path("./.seismometer_cache")
```
diehlbw, line 18: Nit:  Are there cases when the leading ./ is required?
image

src/seismometer/controls/decorators.py
```diff
@@ -2,13 +2,79 @@
...
+    if not SEISMOMETER_CACHE_ENABLED:
+        return func
+
+    @wraps(func)
+    def wrapped_func(*args, **kwargs):
+        arg_str = str((args, frozenset(kwargs.items())))
+        hash_object = hashlib.md5(arg_str.encode())
+        arg_hash = hash_object.hexdigest()
+
+        file_dir = Path("./.seismometer_cache") / "html" / func.__name__
```
diehlbw, line 54: Should this use your global?

tests/controls/test_control_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __str__(self):
+        return self.string
+
+    def __repr__(self):
+        return f"BadReverse({self.string})"
+
+
+class Test_Cached_Function:
+    def teardown_method(self):
```
diehlbw, line 20: This feels more unittest-y   
Is there something that fits better in those patterns than creating a fixture that manipulates state, yields, and performs clean up?  The fixture could include your enabling line too.

src/seismometer/controls/decorators.py
```diff
@@ -2,13 +2,79 @@
...
+from pathlib import Path
 from typing import Any, Callable
 
-from IPython.display import display
+from IPython.display import HTML
 from ipywidgets import Widget
 
+logger = logging.getLogger("seismometer")
+SEISMOMETER_CACHE_DIR = Path("./.seismometer_cache")
+SEISMOMETER_CACHE_ENABLED = os.getenv("SEISMOMETER_CACHE_ENABLED", "") != ""
```
diehlbw, line 19: Can you think of a good place to document how this bit works?  
I think there is some intuition/expectation setting we need around _when_ this occurs, since my read of this is that it's evaluated once at import.

src/seismometer/controls/decorators.py
```diff
@@ -2,13 +2,79 @@
...
+                raise TypeError("The function must return {HTML} object.")
+            file_dir.mkdir(parents=True, exist_ok=True)
+            filepath.write_text(result.data)
+        return result
+
+    return wrapped_func
+
+
+def _clear_cache_dir() -> None:
+    shutil.rmtree(SEISMOMETER_CACHE_DIR / "html", ignore_errors=True)
```
diehlbw, line 72: I'm mostly sold on the extensibility of an html subdirectory, but can you finish convincing me that that is worth not fully cleaning up after ourselves?  
This function removes html but leaves the .seismometer_cache/ dir

src/seismometer/plot/mpl/_ux.py
```diff
@@ -200,11 +199,11 @@ def _register_colormaps() -> None:
     text_cmap = LinearSegmentedColormap.from_list("Text", text_colors, N=len(text_colors))
     alert_cmap = LinearSegmentedColormap.from_list("Alert", alert_colors, N=len(alert_colors))
     neutral_cmap = LinearSegmentedColormap.from_list("Neutral", neutral_colors, N=len(neutral_colors))
-    plt.register_cmap(cmap=area_cmap)
-    plt.register_cmap(cmap=line_cmap)
-    plt.register_cmap(cmap=text_cmap)
-    plt.register_cmap(cmap=alert_cmap)
-    plt.register_cmap(cmap=neutral_cmap)
+    mpl.colormaps.register(cmap=area_cmap)
```
diehlbw, line 202: Adding force=True here will make it safe to run multiple times (at the risk of overwriting someone else's names)

src/seismometer/controls/decorators.py
gbowlin: FINC

src/seismometer/controls/decorators.py
gbowlin: FINC

tests/controls/test_control_decorators.py
gbowlin: Yeah, using a fixture does make more sense

src/seismometer/plot/mpl/_ux.py
gbowlin: FINC

src/seismometer/controls/decorators.py
```diff
@@ -2,13 +2,79 @@
...
+    sig = signature(func)
+    if sig.return_annotation != HTML:
+        raise TypeError("The function must return {HTML} object.")
+
+    if not SEISMOMETER_CACHE_ENABLED:
+        return func
+
+    @wraps(func)
+    def wrapped_func(*args, **kwargs):
+        arg_str = str((args, frozenset(kwargs.items())))
```
diehlbw, line 50: We already chatted but to get it documented, dataframes are annoying.  
the string representation hides rows/columns which can cause collisions + reading parquet doesnt seem to guarantee column order

src/seismometer/controls/decorators.py
gbowlin: You can now clean up the whole cache dir by using DiskCachedFunction.clear_all().

I don't want to clear the cache since we can now re-use it between runs since the hashes for dataframes are now consistent.

tests/controls/test_control_decorators.py
```diff
@@ -0,0 +1,80 @@
...
+        under_test.disk_cached_html_segment.clear_cache()
```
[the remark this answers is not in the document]
gbowlin: updated the tests to use tmp_path and fixtures so we don't have lingering files

src/seismometer/plot/mpl/ux.mplstyle
```diff
@@ -130,12 +130,12 @@ axes.facecolor      : white   # axes background color
...
-#axes.titlesize      : large   # fontsize of the axes title
-#axes.titlepad       : 6.0     # pad between axes and title in points
-#axes.labelsize      : medium  # fontsize of the x any y labels
+axes.titlesize      : large   # fontsize of the axes title
+#axes.titlepad      : 6.0     # pad between axes and title in points
+axes.labelsize      : small  # fontsize of the x any y labels
 #axes.labelpad       : 4.0     # space between label and axis
 #axes.labelweight    : normal  # weight of the x and y labels
-axes.labelcolor     : 221632
+axes.labelcolor     : 5E717A
```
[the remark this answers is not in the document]
gbowlin: This was a request from the UX team - "Field and graph label - Roboto, 13.33pt, Regular (#5E717A)"
I agree its a bit washed out, but it grows on you with our pastel colors.

src/seismometer/controls/decorators.py
gbowlin: updated to have a safe hash based on the dataframe contents.

src/seismometer/controls/decorators.py
gbowlin: I hopefully added it to the sphinx docs under Low Level Patterns.

src/seismometer/controls/decorators.py
diehlbw: I think this was moved making the constants in controls/decorator.py unused. Can they be deleted in this file?  
I'm only seeing usage in the DiskCachedFunction now

src/seismometer/core/decorators.py
```diff
@@ -0,0 +1,232 @@
...
+    Returns
+    -------
+    str
+        hash of the arguments and keyword arguments
+    """
+    hash_object = hashlib.md5()
+    for arg in args:
+        to_hash = _pandas_safe_hash(arg)
+        hash_object.update(str(to_hash).encode())
+    for key, arg in kwargs.items():
```
diehlbw, line 215: Can you put this block undertest?

Also, I think you may need to sort kwargs.  It's good that different combinations of keys+values  are different (line1) but I'd expect different orders of the same keys+values to have the same hash (line2):
image

gbowlin: tests added for sorting, also included some tests around pd.Series (should be sorted by index) and pd.Index (unsorted, as if you are passing an index, the order might be important for caching. )

diehlbw (approved):

PR #30, opened 2024-06-24
Update data load pattern to be extensible

diehlbw:
# Overview

Closes #20 

## Description of changes
Pulled the loading logic for prediction and event data out of the Seismogram class

- [ ] Should maintain functionality for loading from parquet
- [ ] Existing example notebook should work unchanged (ie Verify run_startup works as is)
- [ ] Verify that loading is also possible from preloaded dataframes.  If using the example notebook make sure that the files don't _also_ exist and update run_startup cell to pass in the loaded dataframes.

Note: Updates to allow loading from urls will be done in a follow-up PR

## Author Checklist
- [x] Linting passes; run early with [pre-commit hook](https://pre-commit.com/#install).
- [x] Tests added for new code and issue being fixed.
- [ ] Added type annotations and full numpy-style docstrings for new methods.
- [ ] Draft your news fragment in new `changelog/ISSUE.TYPE.rst` files; see [changelog/README.md](https://github.com/epic-open-source/seismometer/blob/main/changelog/README.md).

PR #34, opened 2024-06-27
Expose memory loader

diehlbw:
# Overview
#30 created a pattern for future data loaders and an option for using pre-loaded dataframes.  
This exposes that functionality to be passed in at setup.

## Description of changes
Add two optional paramaters to run setup.

### Starting Functional Tests:  
Run the example notebook and verify that old pattern works without change.

Turn the urllib.requests cell into a raw or markdown cell (this assumes that the dataset has been pulled locally from the first run)
Add something like the following to a first cell or two :
```python
import seismometer.data.loader as dload
import seismometer.configuration
import seismometer
# Load data from files without calling Seismogram
config = seismometer.configuration.ConfigProvider('.')
loader = dload.loader_factory(config)
data = loader.load_data()

# Now pass that loaded frame through
import pandas as pd
import seismometer as sm
sm.run_startup(config_path='.', predictions_frame=data, events_frame=pd.DataFrame(), log_level=1)
```
Restart and rerun the 

Modifying run_startup revealed some state was unintentionally shared between testcases, so tests were updated to more specifically test the intended logger
## Author Checklist

### DIFF
--- a/src/seismometer/__init__.py
+++ b/src/seismometer/__init__.py
@@ -33,6 +33,9 @@ def run_startup(
     events_frame : Optional[pd.DataFrame], optional
         An optional DataFrame containing the fully loaded events data, by default None.
         By default, when not specified here, these data will be loaded based on conifguration.
+    definitions : Optional[dict], optional
+        A dictionary of definitions to use instead of loading those specified by configuration, by default None.
+        By default, when not specified here, these data will be loaded based on conifguration.
     log_level : logging._Level, optional
         The log level to set. by default, logging.WARN.
     """
--- a/src/seismometer/__init__.py
+++ b/src/seismometer/__init__.py
@@ -16,6 +16,7 @@ def run_startup(
     events_frame: Optional[pd.DataFrame] = None,
     definitions: Optional[dict] = None,
     log_level: int = logging.WARN,
+    reset: bool = False,
 ):
     """
     Runs the required startup for instantiating seismometer.
@@ -38,6 +39,8 @@ def run_startup(
         By default, when not specified here, these data will be loaded based on conifguration.
     log_level : logging._Level, optional
         The log level to set. by default, logging.WARN.
+    reset : bool, optional
+        A flag to completely reset the Seismogram instance, by default False.
     """
     import importlib
 
@@ -51,6 +54,8 @@ def run_startup(
     logger.setLevel(log_level)
     logger.info(f"seismometer version {__version__} starting")
 
+    if reset:
+        Seismogram.kill()
     sg = Seismogram(config_path, output_path, definitions=definitions)
     sg.load_data(predictions=predictions_frame, events=events_frame)
 
--- a/src/seismometer/__init__.py
+++ b/src/seismometer/__init__.py
@@ -40,7 +40,7 @@ def run_startup(
     log_level : logging._Level, optional
         The log level to set. by default, logging.WARN.
     reset : bool, optional
-        A flag to completely reset the Seismogram instance, by default False.
+        A flag when True, will reset the Seismogram instance before loading configuration and data, by default False.
     """
     import importlib
 
--- a/src/seismometer/__init__.py
+++ b/src/seismometer/__init__.py
@@ -1,11 +1,22 @@
 import logging
 from pathlib import Path
+from typing import Optional
+
+import pandas as pd
 
 from seismometer._version import __version__
 from seismometer.core.logger import add_log_formatter, set_default_logger_config
 
 
-def run_startup(*, config_path: str | Path = None, output_path: str | Path = None, log_level: int = logging.WARN):
+def run_startup(
+    *,
+    config_path: str | Path = None,
+    output_path: str | Path = None,
+    predictions_frame: Optional[pd.DataFrame] = None,
+    events_frame: Optional[pd.DataFrame] = None,
+    definitions: Optional[dict] = None,
+    log_level: int = logging.WARN,
+):
     """
     Runs the required startup for instantiating seismometer.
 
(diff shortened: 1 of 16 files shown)


src/seismometer/__init__.py
```diff
@@ -16,6 +27,12 @@ def run_startup(*, config_path: str | Path = None, output_path: str | Path = Non
     output_path : Optional[str | Path], optional
         An output path to write data to, overwriting the default path specified by info_dir in config.yml,
         by default None.
+    predictions_frame : Optional[pd.DataFrame], optional
+        An optional DataFrame containing the fully loaded predictions data, by default None.
+        By default, when not specified here, these data will be loaded based on conifguration.
+    events_frame : Optional[pd.DataFrame], optional
+        An optional DataFrame containing the fully loaded events data, by default None.
+        By default, when not specified here, these data will be loaded based on conifguration.
```
gbowlin, line 35: Missing a comment for 
    definitions: Optional[dict] = None,

```
definitions : Optional[dict], optional
        A dictionary of definitions to use instead of loading those specified by configuration, by default None.
```

gbowlin (commented):
missing comment in header

src/seismometer/__init__.py
diehlbw: FINC

src/seismometer/seismogram.py
```diff
@@ -73,13 +80,36 @@ def __init__(self, config_path: Optional[str | Path] = None, output_path: Option
...
+        predictions : pd.DataFrame, optional
+            The fully prepared predictions dataframe, by default None.
+            Uses this when specified, otherwise loads based on configuration.
+        events : pd.DataFrame, optional
+            The pre-loaded events dataframe, by default None.
+            Uses this when specified, otherwise loads based on configuration.
+        reset : bool, optional
+            Flag when set to true will overwrite existing dataframe, by default False
+        """
+        if self.dataframe and not reset:
```
gbowlin, line 109: self.dataframe has not been defined yet, so self._dataframe does not exist, 

You can fix this by adding 

```python
        self.dataframe: pd.DataFrame = None
```

to `__init__`

src/seismometer/__init__.py
```diff
@@ -1,11 +1,22 @@
...
-def run_startup(*, config_path: str | Path = None, output_path: str | Path = None, log_level: int = logging.WARN):
+def run_startup(
+    *,
+    config_path: str | Path = None,
+    output_path: str | Path = None,
+    predictions_frame: Optional[pd.DataFrame] = None,
+    events_frame: Optional[pd.DataFrame] = None,
+    definitions: Optional[dict] = None,
+    log_level: int = logging.WARN,
+):
```
gbowlin, line 19: For ease of testing, you should probably include a reset flag here too

src/seismometer/__init__.py
```diff
@@ -31,8 +51,8 @@ def run_startup(*, config_path: str | Path = None, output_path: str | Path = Non
     logger.setLevel(log_level)
     logger.info(f"seismometer version {__version__} starting")
 
-    sg = Seismogram(config_path, output_path)
-    sg.load_data()
+    sg = Seismogram(config_path, output_path, definitions=definitions)
+    sg.load_data(predictions=predictions_frame, events=events_frame)
```
gbowlin, line 55: passthrough a reset flag here to make it easy to reload in the classifier_bin.ipynb

src/seismometer/seismogram.py
```diff
@@ -73,13 +80,36 @@ def __init__(self, config_path: Optional[str | Path] = None, output_path: Option
...
+            logger.debug("Data already loaded; pass reset=True to clear data and re-evaluate.")
+            return
+
         self._load_metadata()
 
         self.dataframe = self.dataloader.load_data(predictions, events)
```
gbowlin, line 115: Do we need any special handling when events is none but predictions is not none?

src/seismometer/__init__.py
```diff
@@ -1,11 +1,22 @@
...
 from seismometer.core.logger import add_log_formatter, set_default_logger_config
 
 
-def run_startup(*, config_path: str | Path = None, output_path: str | Path = None, log_level: int = logging.WARN):
+def run_startup(
+    *,
+    config_path: str | Path = None,
+    output_path: str | Path = None,
+    predictions_frame: Optional[pd.DataFrame] = None,
+    events_frame: Optional[pd.DataFrame] = None,
```
gbowlin, line 16:
```python
pred_df = pd.read_parquet("data/predictions.parquet")
event_df = pd.read_parquet("data/events.parquet")


import seismometer as sm
sm.run_startup(config_path='.', predictions_frame=pred_df, events_frame=event_df, log_level=1)
```

Fails because the dataframe for events uses "EventTime" over "Time"

Is that expected, or should we still listen to the mapping done in Config?

gbowlin: Doing the mapping myself does fix the issues

```python
import pandas as pd
pred_df = pd.read_parquet("data/predictions.parquet")
event_df = pd.read_parquet("data/events.parquet")
event_df.rename({"EventTime": "Time"}, inplace=True, axis=1)


import seismometer as sm
sm.run_startup(config_path='.', predictions_frame=pred_df, events_frame=event_df, log_level=1)
```

This does work

gbowlin (changes requested):
The major question I have is around if we should apply mappings to the data-frames we are handed. We do handling age binning, but we don't remap columns if the event dataframe's Time column is not the expected column.

src/seismometer/__init__.py
diehlbw: I think this is the most consistent expectation, that passed frames should be ready for usage.  
There is room to improve the warning messages to make this more apparent, as the transforms after load are likely not known.

src/seismometer/seismogram.py
diehlbw: I think the pipeline._add_events code actually handles this sufficiently.

My guess is you hit the same thing I did when trying to test and see what you meant, which is that an events.parquet exists and gets loaded when we specify None (or omit the kwarg).  
So for testing, delete that file and verify the case is supported as you'd want.  

Additionally, let's write up something to more robustly handle the case when predictions has the *_Value and/or *_Time columns that merging wants to add. 
My instinct is to prefer prediction data, and just warn about skipping the merge.

gbowlin: Can you write the issue up, then I'll resolve this.

src/seismometer/__init__.py
gbowlin: Can you write up an issue to make the error messages better, more descriptive, suggestive of potential fixes?

Then I'll mark this as resolved

gbowlin (approved):

src/seismometer/__init__.py
diehlbw: Wrote up the issue in #39

src/seismometer/seismogram.py
diehlbw: Wrote up the issue in #39

PR #41, opened 2024-07-02
Update Plotting Controls to allow Exploration of Settings

gbowlin:
# Overview

This is a Draft to get some developer feedback on some larger changes around exploration of model data.
 
This is a pretty big change to allow model exploration by wrapping plots in user friendly controls rather than setting values directly on the Seismogram Singleton. 

## Author Checklist


diehlbw (changes requested):

changelog/41.feature.rst
```diff
@@ -0,0 +1 @@
+Added Exploration controls to allow setting dyanmic values for thresholds, targets, outcomes, etc.
```
diehlbw, line 1:
```suggestion
Added Exploration controls to allow setting dynamic values for thresholds, targets, outcomes, etc.
```

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,916 @@
...
+        @wraps(callback)
+        def callback_wrapper(button):
+            button.description = self.UPDATING_PLOTS
+            button.disabled = True
+            callback(button)
+            button.description = self.UPDATE_PLOTS
+
+        self.plot_button.on_click(callback_wrapper)
+
+    def on_toggle_code(self, callback):
```
diehlbw, line 63: Should toggling the code checkbox enable the update button?

src/seismometer/controls/cohort_comparison.py
```diff
@@ -21,15 +21,23 @@ def __init__(self, selections: dict[str, tuple[any]], exclude_cols: Optional[lis
...
         self.output = Output()
         self.button = Button(description=GENERATE_REPORT, button_style="primary")
         self.button.on_click(partial(self._generate_comparison_report, self))
 
     def show(self):
-        display(VBox(children=[HBox(children=self.selectors), self.button, self.output]))
+        display(
+            VBox(
+                children=[
+                    Box(children=self.selectors, layout=Layout(grid_gap="20px", align_times="flex-start")),
```
diehlbw, line 35: What is align_times?  
If this is align_items, you could use BOX_GRID_LAYOUT

changelog/41.feature.rst
gbowlin: FINC

src/seismometer/controls/cohort_comparison.py
gbowlin: Spelling at its worst, this should be BOX_GRID_LAYOUT. 

FINC

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,140 @@
...
+        Parameters
+        ----------
+        names : tuple[str]
+            Slider names
+        value : Optional[tuple[str]], optional
+            Slider start values, by default None, starts all sliders at zero.
+        """
+        super().__init__()
+        self.names = tuple(names)  # make immutable
+        self.value = value or tuple(0 for _ in names)  # set initial value
```
diehlbw, line 30: Should initialization do more validation on passed values?  It's weird that untouched controls can have a state that is not possible to return to
image
As soon as you touch it, does get pruned/winsored: ex (0.88, 1.0)

src/seismometer/data/filter.py
```diff
@@ -361,6 +361,34 @@ def none(cls):
...
         return cls(None, "none")
 
+    @classmethod
+    def from_cohort_dictionary(cls, cohort_dict: dict[str, tuple[any]] | None = None) -> "FilterRule":
+        """
+        For a given dictionary, generate a matching FilterRule
+
+        Parameters
+        ----------
+        cohort_dict : dict[str,tuplep[any]], optional
```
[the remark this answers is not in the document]
gbowlin: FINC - found the same typo elsewhere

src/seismometer/controls/selection.py
```diff
@@ -39,52 +38,44 @@ def __init__(self, title: str, options: tuple[str], value: tuple[str] = None, sh
...
+            self.button_from_option[option].value = value
 
     def get_selection_text(self) -> str:
-        """Descriptoin of the currently selected values."""
-        if self.control_value:
-            return f"{self.title}: {','.join([str(x) for x in self.control_value])}"
+        """Description of the currently selected values."""
+        text = f"{self.title_label.value}: " if self.title_label else ""
+        if self.value:
+            return text + f"{','.join([str(x) for x in self.value])}"
```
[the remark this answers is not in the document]
gbowlin: its human only, so I'll add spaces

src/seismometer/controls/selection.py
```diff
@@ -205,39 +209,38 @@ def __init__(
...
-    @control_value.setter
-    def control_value(self, new_value):
-        """Update all the children based on the new value."""
-        key = new_value[0]
-        value = new_value[1]
-        self.selection_widgets[key].control_value = value
+        self.value = (key, tuple(selections))
+
+    def _on_value_change(self, change=None):
+        """Buble up changes."""
```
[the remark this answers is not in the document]
gbowlin: FINC

src/seismometer/controls/thresholds.py
gbowlin: Added initial clamping, and some validation around values being in the correct order (for Monotonic Sliders)

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,148 @@
...
+        value : Optional[tuple[str]], optional
+            Slider start values, by default None, starts all sliders at zero.
+        increasing : bool, optional = True
+            Forces sliders to be increasing, else decreasing.
+            If initial values are not sorted, raise an ValueError.
+        """
+        super().__init__(names, value)
+
+        if tuple(sorted(self.value.values(), reverse=not increasing)) != tuple(self.value.values()):
+            raise ValueError("Initial values are not sorted")
```
diehlbw, line 110: Can you add expected direction to the error message?  They might be "sorted" but just need the increasing flag toggled.

Nit. pandas uses the flag ascending rather than increasing, should we align? (numpy doesn't seem to even expose the option)

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,148 @@
...
+from seismometer.plot.mpl._ux import alert_colors
+
+from .styles import WIDE_LABEL_STYLE
+
+logger = logging.getLogger("seismometer")
+
+
+class PercentSliderListWidget(ValueWidget, VBox):
+    """
+    Vertical list of sliders
```
diehlbw, line 16: Can you add to the docstring that these are 0-1 in increments of 0.01.  
Two things that I want to make explicit. The bounds (-250% is well defined, I think, but not allowed) and that your range are NOT the integer percentiles (0-100).  

Elsewhere for logit 'scores' I believe we have used 'percentiles' as 0-100 and 'probabilities' as 0-1.

src/seismometer/controls/thresholds.py
gbowlin: moving to ascending to match pandas (since python's reverse feels less intuitive)

FINC

src/seismometer/controls/thresholds.py
gbowlin: Will do, also changing the name to ProbabilitySliderListWidget. I'll write up an issue for controls needing to support non 0-1 thresholds and needing threshold information in config.yml to allow named threshold and threshold limits in the future.

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,1109 @@
...
+        super().__init__(
+            "Outcome / Intervention Analysis",
+            sg.available_cohort_groups,
+            tuple(sg.config.outcomes.keys()),
+            tuple(sg.config.interventions.keys()),
+            reference_times,
+        )
+
+    def generate_plot(self) -> HTML:
+        from seismometer._api import plot_intervention_outcome_timeseries
```
diehlbw, line 1086: So if we generalize that fn.__name__ bit, it feels like there is a way to shift the seam of controls to be only the options (plus plot signature).

I haven't looked completely but I'd hope we can move plotting imports completely out of the control submodule; instead _api can import controls and construct the callable instances having specified that last bit of info.

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,1109 @@
...
+        def callback_wrapper(change):
+            callback(self.code_checkbox.value)
+
+        self.code_checkbox.observe(callback_wrapper, "value")
+
+    def trigger(self):
+        self.plot_button.click()
+
+
+class ExlorationWidget(VBox):
```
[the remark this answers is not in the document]
gbowlin: FINC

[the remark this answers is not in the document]
gbowlin: FINC

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,1109 @@
...
+                for x in [
+                    self.option_widget.cohort,
+                    self.option_widget.cohort_groups,
+                    self.option_widget.target,
+                    self.option_widget.score,
+                    list(self.option_widget.thresholds.values()),
+                ]
+            ]
+        )
+        return f"sm.plot_cohort_evaluation({args}, per_context={self.option_widget.group_scores})"
```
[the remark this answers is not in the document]
gbowlin: It was a bit of effort, but well worth it, now the Exploration* classes implement a `generate_plot_args` method, so that we can just pass in the actual plot function. I've added an example of this to the `classifier_bin` notebook. 

This also allows me to move all the custom Expore* objects directly into the `_api` module so cleanup startup some.

src/seismometer/controls/explore.py
gbowlin: FINC

tests/controls/test_control_decorators.py
```diff
@@ -0,0 +1,11 @@
+from IPython.display import HTML
+
+import seismometer.controls.decorators as undertest
+
+
+class TestCachedHTML:
+    def test_save_load_html(self, tmp_path):
+        html_path = tmp_path / "test.thml"
```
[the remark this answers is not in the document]
gbowlin: FINC

src/seismometer/_api.py
```diff
@@ -1058,7 +1243,157 @@ def show_cohort_summaries(by_target: bool = False, by_score: bool = False):
...
+    def __init__(self):
+        """
+        Exploration widget for model evaluation.
+
+        This includes the ROC, recall vs predicted condition prevalence, calibration,
+        PPV vs sensitivity, sensitivity/specificity/ppv, and a histogram.
+        """
+        from seismometer._api import plot_model_evaluation
+
+        super().__init__("Model Performance", plot_model_evaluation)
```
diehlbw, line 1269: Question: Perhaps we should have a sidebar on the goal for how these are interacted with (I'm fine not holding this PR up over this as I think a couple followups are already expected), but with the simplicity here, I think there are options around removal of this child class completely and doing something more direct like:  
`explore_model_evaluation = ExplorationModelSubgroupEvaluationWidget("Model Performance", plot_model_evaluation)`

And at that point, we could still "act" like a function by adding a `__call__: return self` if we don't like the look of jupyter cells that are recalling.  
I suppose the real decision-point is that a single instance means multiple cells can become linked, for better or worse.

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,907 @@
...
+    def group_scores(self) -> bool:
+        """If scores should be grouped by context"""
+        return self.model_options.group_scores
+
+
+# endregion
+# region Exploration Widgets (Intermediate Classes)
+
+
+class ExplorationWidget(VBox):
```
diehlbw, line 605: Nit: is this really "Intermediate"?  It's still based on VBox, and the other ones in this region are mostly based on this one

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,907 @@
...
+                cohort_subgroups: tuple[Any]
+                target: str,
+                score: str,
+                thresholds: tuple[float],
+                *, per_context: bool) -> Any
+        """
+        from seismometer.seismogram import Seismogram
+
+        sg = Seismogram()
+        match theshold_handling:
```
diehlbw, line 801:
```suggestion
        match threshold_handling:
```

src/seismometer/controls/explore.py
```diff
@@ -0,0 +1,907 @@
...
+            case "all":
+                self.thresholds = {f"Threshold {k}": v for k, v in enumerate(sorted(sg.thresholds, reverse=True), 1)}
+            case "max":
+                self.thresholds = {"Threshold": max(sg.thresholds)}
+            case "min":
+                self.thresholds = {"Threshold": min(sg.thresholds)}
+            case _:
+                self.thresholds = None
+
+        self.threshold_hanlding = theshold_handling
```
diehlbw, line 811:
```suggestion
        self.threshold_handling = threshold_handling
```
handling (on left), threshold (on right)  
I think this was also copied into some other classes

docs/reference/seismometer.rst
```diff
@@ -9,9 +9,14 @@ Public API
 .. autosummary::
    :toctree: api/
 
+   ExploreModelEvaluation
```
diehlbw, line 12: Consideration:  I almost think these deserve their own section.  This is becoming _the_ public api more than the old functions.  
Could also be a followup PR, as I think we'll want some dedicated guide section(s) the control+plot pattern

src/seismometer/_api.py
```diff
@@ -1058,7 +1243,157 @@ def show_cohort_summaries(by_target: bool = False, by_score: bool = False):
...
+
+
+@export
+class ExploreCohortEvaluation(ExplorationCohortSubclassEvaluationWidget):
+    """
+    A widget for exploring the model performance of a cohort.
```
diehlbw, line 1275: Can we get this first line of the docstring to be a little more specific; the goal being to help a user know which they want to use from the sphinx TOC:
image

And note that the table pictured only displays that first-line summary. On following the link you then see the anything beyond the first blankline of the class docstring (and of course the constructor, methods, etc)

src/seismometer/_api.py
diehlbw: neverminding: the "linking" is a problem.  the class is a good way to avoid it.

src/seismometer/_api.py
```diff
@@ -1058,7 +1243,157 @@ def show_cohort_summaries(by_target: bool = False, by_score: bool = False):
...
+    """
+
+    def __init__(self):
+        """
+        Exploration widget for model evaluation.
+
+        This includes the ROC, recall vs predicted condition prevalence, calibration,
+        PPV vs sensitivity, sensitivity/specificity/ppv, and a histogram.
+        """
+        from seismometer._api import plot_model_evaluation
```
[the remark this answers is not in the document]
gbowlin: FINC

src/seismometer/controls/explore.py
gbowlin: FINC

src/seismometer/controls/explore.py
gbowlin: FINC

src/seismometer/controls/explore.py
gbowlin: FINC

src/seismometer/_api.py
gbowlin: image

Thanks for the suggestion!

docs/reference/seismometer.rst
gbowlin: FINC

diehlbw (approved):

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,154 @@
...
+        ascending : bool, optional = True
+            Forces sliders to be ascending, else decreasing.
+            If initial values are not sorted, raise an ValueError.
+        """
+        if len(names) > 6:
+            raise ValueError("MonotonicProbabilitySliderListWidget only supports up to 6 sliders")
+
+        super().__init__(names, value)
+
+        if tuple(sorted(self.value.values(), reverse=not ascending)) != tuple(self.value.values()):
```
andli28, line 113: `self.value.values()` inplies that `self.value` is expected to be a dictionary, but the function signature suggests its a tuple.

src/seismometer/html/iframe.py
```diff
@@ -0,0 +1,7 @@
+from pathlib import Path
+
+from IPython.display import IFrame
+
+
+def load_as_iframe(path: Path, height: int = 300) -> IFrame:
```
gbowlin, line 6: While this is good for caching, and makes sure that any javascript for the new html segment gets run, it does have two drawbacks:

1. VSCode does not support loading an IFrame source from disk - https://github.com/microsoft/vscode/issues/154722
2. JHub requires some handling for this to propagate to jupyter lab instances - https://stackoverflow.com/questions/60257156/not-able-to-embed-jupyterlab-running-on-jupyterhub-into-an-iframe

src/seismometer/controls/thresholds.py
andli28, line 113: Also, wouldn't it be more efficient to do a linear pass through the tuple rather than sorting and comparing?

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,154 @@
...
+        """Slider has changed, update the value tuple, making sure values are increating"""
+        sliders = list(self.sliders.values())
+        if self.value_update_in_progress:
+            return
+        try:
+            slider_index = sliders.index(change["owner"])
+        except ValueError:  # Only the sliders can change values
+            return
+        new = change["new"]
+        old = change["old"]
```
andli28, line 135: If change is default None, would this fail?

If change["owner"] exists but change{'new"] nad change["old"] doesn't, would this fail?

andli28 (changes requested):
Review for thresholds,py

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,154 @@
...
+        self.ascending = ascending
+        for index, sub_slider in enumerate(self.sliders.values()):
+            if ascending:
+                # line them up so that last slider is the first alert color
+                sub_slider.style.handle_color = alert_colors[(len(self.value) - index - 1) % len(alert_colors)]
+            else:
+                sub_slider.style.handle_color = alert_colors[index % len(alert_colors)]
+
+    def _on_slider_change(self, change=None):
+        """Slider has changed, update the value tuple, making sure values are increating"""
```
andli28, line 126: increating -> increasing

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,154 @@
...
+        names : tuple[str]
+            Slider names
+        value : Optional[tuple[str]], optional
+            Slider start values, by default None, starts all sliders at zero.
+        ascending : bool, optional = True
+            Forces sliders to be ascending, else decreasing.
+            If initial values are not sorted, raise an ValueError.
+        """
+        if len(names) > 6:
+            raise ValueError("MonotonicProbabilitySliderListWidget only supports up to 6 sliders")
```
andli28, line 109: I can't figure out why, but Pylance is warning that this line of Code is unreachable. Might be a good idea to test this in test_thresholds.py

andli28: Also just realized that `test_thesholds.py` should be spelled `test_thresholds.py`

andli28:
`test_thesholds.py` should be spelled `test_thresholds.py`

src/seismometer/controls/selection.py
```diff
@@ -11,26 +11,25 @@ class SelectionListWidget(ValueWidget, VBox):
...
         ----------
-        title : str
-            Title to be displayed above the list of buttons.
         options : tuple[str]
             Selectable options.
         value : Optional[tuple[str]], optional
             Subset of options that should be selected by default, by default None.
-        show_title : bool, optional
-            If True, shows the title, by default True.
+        title : Optional[str[], optoinal
```
andli28, line 23: optoinal -> optional

andli28: there should be a spellchecker extension in vs code 😆

src/seismometer/controls/selection.py
```diff
@@ -11,26 +11,25 @@ class SelectionListWidget(ValueWidget, VBox):
...
+            Title to be displayed above the list of buttons, it not set, title not included
         """
         super().__init__()
         self.options = tuple(options)  # make immutable
         selected_values = value or ()
         selected_values = tuple(value for value in selected_values)
```
andli28, line 29: This looks redundant, since selected_values is already a tuple.

src/seismometer/controls/selection.py
```diff
@@ -11,26 +11,25 @@ class SelectionListWidget(ValueWidget, VBox):
...
-        title : str
-            Title to be displayed above the list of buttons.
         options : tuple[str]
             Selectable options.
         value : Optional[tuple[str]], optional
             Subset of options that should be selected by default, by default None.
-        show_title : bool, optional
-            If True, shows the title, by default True.
+        title : Optional[str[], optoinal
+            Title to be displayed above the list of buttons, it not set, title not included
```
andli28, line 24: it not set, title not included -> **if** not set, title not included

src/seismometer/controls/selection.py
```diff
@@ -112,59 +114,83 @@ def __init__(
...
-    def control_value(self, new_value):
-        """Updates the actual control value."""
-        for key in self.selection_widgets:
-            self.selection_widgets[key].control_value = new_value.get(key, ())
+    def _on_value_change(self, change=None):
+        """Bubble up changes."""
+        self.value_update_in_progress = True
+        updated_values = {key: self.value.get(key, ()) for key in self.selection_widgets}
+        for key, value in updated_values.items():
+            self.selection_widgets[key].value = self.value.get(key, ())
```
andli28, line 177: replacing `self.value.get(key, ())` with `value` 

or

```for key in self.selection_widgets:
            self.selection_widgets[key].value = self.value.get(key, ()) ```

would be more efficient I believe

File: src/seismometer/controls/selection.py [hunk 30]
```diff
@@ -181,7 +207,7 @@ def __init__(
         value: Optional[tuple[str | tuple]] = None,
```
andli28, line 207: I think a comma instead of a "|" is appropriate

src/seismometer/controls/selection.py
```diff
@@ -196,7 +222,7 @@ def __init__(
         title : str, optional
             Dispaly above the dropdown, by default "".
```
andli28, line 223: Display

src/seismometer/controls/selection.py
```diff
@@ -181,7 +207,7 @@ def __init__(
         value: Optional[tuple[str | tuple]] = None,
         *,
         title: str = "",
-        select_all: bool = False,
+        select_all: bool = True,
     ):
```
andli28, line 211: Can't comment on line 219, but drown should be dropdown in docstring

src/seismometer/controls/selection.py
```diff
@@ -205,39 +231,50 @@ def __init__(
...
-                title=key, options=options[key], value=values[key], show_title=False
-            )
+            selection_widget = SelectionListWidget(options=options[key], value=values[key])
             self.selection_widgets[key] = selection_widget
             selection_widget.observe(self._on_selection_change, "value")
         self.stack = Stack(children=[self.selection_widgets[key] for key in self.selection_widgets], selected_index=0)
         self.children = [self.label, self.dropdown, self.stack]
         jslink((self.dropdown, "index"), (self.stack, "selected_index"))
-        self.layout = Layout(width="max-content")
+        self.layout = Layout(width="calc(100& - var(--jp-widgets-border-width)* 2)")
```
andli28, line 249: Should 100& be 100%?

src/seismometer/controls/selection.py
```diff
@@ -205,39 +231,50 @@ def __init__(
         if value is not None:
             values[value[0]] = value[1]
         else:
-            value = (list(values.keys())[0], values[list(values.keys())[0]])
+            value = (tuple(values.keys())[0], values[tuple(values.keys())[0]])
```
andli28, line 234: value initializing seems very convoluted - note to self to investigate further

src/seismometer/html/iframe.py
gbowlin: Written up as https://github.com/epic-open-source/seismometer/issues/53

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,154 @@
...
+    value = traitlets.Dict(help="The names and values for the slider list")
+
+    def __init__(self, names: Iterable[str], value: Iterable[int] = None):
+        """A vertical list of sliders, bounded between 0 and 1 with a step of 0.01.
+
+        Parameters
+        ----------
+        names : tuple[str]
+            Slider names
+        value : Optional[tuple[str]], optional
```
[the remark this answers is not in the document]
gbowlin: FINC

src/seismometer/controls/thresholds.py
```diff
@@ -0,0 +1,154 @@
...
+            Slider start values, by default None, starts all sliders at zero.
+        ascending : bool, optional = True
+            Forces sliders to be ascending, else decreasing.
+            If initial values are not sorted, raise an ValueError.
```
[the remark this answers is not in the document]
gbowlin: Not intentional, it came from testing ways to handle getting a set of names and thresholds, where sets are Iterable, but unordered. We decided that handling sets was not needed, but keeping Iterable as a type hint allows tuples and lists.

src/seismometer/controls/thresholds.py
gbowlin: This is self.values, which is a dict (see the parent class ProbabilitySliderListWidget). 

super().__init__(names, values) creates self.values which is a dict of threshold names to threshold values_

gbowlin: Yes, but this is a pretty rare code path, with only a few values to sort, so not that slow. that being said, I did do this change mostly because I don't need to call sorted and have the casts to tuple. 

FINC

src/seismometer/controls/thresholds.py
gbowlin: If owner exists, we will have a new and old value.

src/seismometer/controls/thresholds.py
gbowlin: FINC

src/seismometer/controls/thresholds.py
gbowlin: FINC, also see `test_too_many_thresholds_raises_value_error` which tests that line (my Pylance when from warning to working after I fixed the type hint for names/values)

src/seismometer/controls/selection.py
gbowlin: There is, and I should be a better user of it. Under `Problems` you can see spelling issues

image

FINC - and I ran through all the spelling issues from the problem list.

src/seismometer/controls/selection.py
gbowlin: It could be something else like a list, while the type hint asks for a tuple, this will work with lists, I made this more clear by adopting the Iterable type hint.

src/seismometer/controls/selection.py
gbowlin: FINC

src/seismometer/controls/selection.py
gbowlin: But we also want to freeze the values of all the widgets when the request first comes in. As we loop over values, we end up changing them. 

It shouldn't matter since we end up skipping reverse updates in `_on_subselection_change` while this is happening, I wanted to make sure I only have an immutable object while I make my changes, rather than `self.value` which is mutable.

src/seismometer/controls/selection.py
gbowlin: Yes, FINC

src/seismometer/controls/selection.py
gbowlin: FINC

src/seismometer/controls/selection.py
gbowlin: added some comments, and re-wrote the code to hopefully make this more transparent.

src/seismometer/_api.py
```diff
@@ -23,7 +31,8 @@
 from .data.filter import FilterRule
 from .data.timeseries import create_metric_timeseries
 from .html import template
-from .report.auditing import display_fairness_audit
+from .html.iframe import load_as_iframe
+from .report.auditing import fairness_audit_to_html
 from .report.profiling import ComparisonReportWrapper, SingleReportWrapper
 from .seismogram import Seismogram
```
andli28, line 38: This refers to the typehint for function `feature_alerts`: should be `Optional[list[str]] = None`

Don't know why Github doesn't support commenting on collapsed lines.

andli28, line 38: Line 76: This refers to the typehint for `exclude_cols`: should be Optional[list[str]] = None

Maybe this is fine, but it would be inconsistent with typehints in `thresholds.py`

src/seismometer/_api.py
```diff
@@ -148,14 +157,12 @@ def target_feature_summary(exclude_cols: list[str] = None, inline=False):
 
 
 @export
-def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1.25) -> None:
+def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1.25) -> HTML:
```
andli28, line 160: same thing: typehint for `metric_list` does not match docstring

src/seismometer/_api.py
```diff
@@ -176,86 +183,126 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
+        metric_list=metric_list,
+        fairness_threshold=fairness_threshold,
     )
 
 
-# endregion
-
-# region notebook IPWidgets # TODO - move to .controls submodule
+@export
+def generate_fairness_audit(
```
andli28, line 198: Q: when are docstrings needed and when are they not?

src/seismometer/_api.py
```diff
@@ -176,86 +183,126 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
-    def on_widget_value_changed(*args):
-        sg.selected_cohort = widget.value
 
-    widget.observe(on_widget_value_changed, "value")
+# endregion
 
-    # get intial value
-    on_widget_value_changed()
-    return widget
+# region notebook IPWidgets # TODO - move to .controls submodule
```
andli28, line 248: Is this still a TODO?

src/seismometer/_api.py
andli28, line 160: for consistency, would it make sense to have a typehint for fairness_threshold despite it being assigned ot a float? Something like `fairness_threshold: float = 1.25`

src/seismometer/_api.py
```diff
@@ -176,86 +183,126 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
-        aggregation_method=sg.event_aggregation_method(sg.target),
-    )[[sg.target, sg.output] + sensitive_groups]
-
-    display_fairness_audit(
-        df, sensitive_groups, sg.output, sg.target, sg.thresholds[0], metric_list, fairness_threshold
+    return generate_fairness_audit(
+        sensitive_groups,
+        sg.target,
+        sg.output,
+        sg.thresholds[0],
```
andli28, line 190: Q: I see that thresholds is a list of floats. What is the significance of thresholds[0]?

src/seismometer/_api.py
```diff
@@ -176,86 +183,126 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
 
     options = sg.available_cohort_groups
 
-    comparison_selections = MultiSelectionListWidget(
-        options, ghost_text="Select cohort groups", title="Available Cohorts"
-    )
+    comparison_selections = MultiSelectionListWidget(options, title="Cohort")
     output = Output()
 
     def on_widget_value_changed(*args):
```
andli28, line 266: *args isn't being used - is this still a necessary parameter?

src/seismometer/_api.py
```diff
@@ -176,86 +183,126 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
-                rule = filter_rule_from_cohort_dictionary(comparison_selections.value)
-                cohort_count = sum(rule.mask(sg.dataframe))
-                if cohort_count < 10:
-                    print("Selected Cohort is empty, or has fewer than 10 predictions.")
-                else:
-                    print(f"Selected Cohort has {sum(rule.mask(sg.dataframe))} predictions.")
-            else:
-                print(f"No Cohort Selected: {len(sg.dataframe)} predictions total.")
+            html = _cohort_list_details(comparison_selections.value)
+            output.clear_output()
```
andli28, line 269: Should we clear the output before generating new content? I understand that we want to make the transition seem smooth, but if generation takes a long time it will seem like nothing is happening to the user, which is frustrating. I would argue with a cleared output at least the user knows that the update is in progress.

It really comes down to how long we expect generating new content to be. If its quick then this is fine. Otherwise I think its better to clear first.

src/seismometer/_api.py
```diff
@@ -176,86 +183,126 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
+    rule = filter_rule_from_cohort_dictionary(cohort_dict)
+    data = rule.filter(sg.dataframe)
+    cohort_count = data[sg.entity_keys[0]].nunique()
+    if cohort_count < sg.censor_threshold:
+        return template.render_censored_plot_message(sg.censor_threshold)
+
+    cfg = sg.config
+    target_cols = [pdh.event_value(x) for x in cfg.targets]
+    intervention_cols = [pdh.event_value(x) for x in cfg.interventions]
+    outcome_cols = [pdh.event_value(x) for x in cfg.outcomes]
```
andli28, line 294: in pandas_helpers: event_value, event_time, event_name: Pylance is saying the first `return None` is unreachable, because the typehints say that event must be a str.

`def event_value(event: Optional[str]) -> Optional[str]:` would probably be more accurate

src/seismometer/_api.py
gbowlin: its the first threshold in metadata.json, so potentially the most important? Since the list is unordered, its kind of hard to tell, but in most cases we have just said 1st = primary. This is why the explore controls allow you to change this.

src/seismometer/_api.py
gbowlin: yes, because the caller still sends us args, even if we ignore them.

src/seismometer/_api.py
gbowlin: We should try to include them for anything we expect to be a public facing API.  The most common exception should be private methods (prefixed with an underscore) 

A:  This should have a docstring :-)

src/seismometer/_api.py
gbowlin: Probably because they weren't modified, but I agree it would be nice if you could.

gbowlin: FINC

src/seismometer/_api.py
gbowlin: FINC

src/seismometer/_api.py
gbowlin: FINC

src/seismometer/_api.py
gbowlin: yes and no. 

No in the sense that I fixed up what I intended to in this PR, but in general it would be nice to have _api.py done differently. 

Removing comment.

src/seismometer/_api.py
gbowlin: FINC

src/seismometer/_api.py
gbowlin: Its currently used in the two older plotting functions (`plot_leadtime_enc` and `fairness_audit`), which will be removed in a future PR because we now have the Explore* controls that will replace them

src/seismometer/_api.py
gbowlin: Added a "Recalculating..." message

src/seismometer/_api.py
gbowlin: Instead removed the None check, since the event is typed to string.

src/seismometer/_api.py
```diff
@@ -269,13 +344,28 @@ def plot_cohort_hist():
     cohort_col = sg.selected_cohort[0]
     subgroups = sg.selected_cohort[1]
     censor_threshold = sg.censor_threshold
-    return _plot_cohort_hist(sg.data(), sg.target_event, sg.target, sg.output, cohort_col, subgroups, censor_threshold)
+    return _plot_cohort_hist(sg.data(), sg.target, sg.output, cohort_col, subgroups, censor_threshold)
+
+
+@disk_cached_html_segment
+@export
+def plot_cohort_group_histograms(cohort_col: str, subgroups: list[str], target_column: str, score_column: str) -> HTML:
```
andli28, line 352: Docstring needed?

src/seismometer/_api.py
```diff
@@ -307,29 +395,27 @@ def _plot_cohort_hist(
...
 
     if len(cData.index) == 0:
-        logger.error(
-            "No groups were left uncensored; timeframe is too small in combination with frequency of "
-            + f"{target_event} and the number of cohorts"
-        )
-        return
+        return template.render_censored_plot_message(censor_threshold)
 
     bins = np.histogram_bin_edges(cData["pred"], bins=20)
```
andli28, line 408: What's special about magic number 20? Should this be a parameter in the function or a constant defined somewhere?

gbowlin: 20 is just a nice number of bins

andli28:
Running `pylint src/seismometer` shows some cyclic-imports - is this something to be concerned about?

andli28 (changes requested):
Running `pylint src/seismometer` shows some cyclic-imports - is this something to be concerned about?

pylint also reveals places where there's implicit str concat, inconsistent return expressions, and "consider-using-f-string" and "consider-using-from-import", "logging-not-lazy"... too many to list here.

Many of the warnings aren't a concern in my opinion like long lines, but some may be

src/seismometer/controls/selection.py
```diff
@@ -11,26 +12,25 @@ class SelectionListWidget(ValueWidget, VBox):
...
         ----------
-        title : str
-            Title to be displayed above the list of buttons.
-        options : tuple[str]
+        options : Iterable[str]
             Selectable options.
-        value : Optional[tuple[str]], optional
+        value : Optional[Iterable[str]], optional
             Subset of options that should be selected by default, by default None.
-        show_title : bool, optional
```
andli28, line 25: missing ]

src/seismometer/controls/selection.py
```diff
@@ -39,57 +39,60 @@ def __init__(self, title: str, options: tuple[str], value: tuple[str] = None, sh
...
+            self.button_from_option[option] = sub_toggle
+            self.option_from_button[sub_toggle] = option
 
-        self.layout = Layout(width="max-content", min_width="100px")
-        self.label = Label(title)
-        if show_title:
-            self.children = [self.label] + self.buttons
+        self.layout = Layout(width="max-content", min_width="var(--jp-widgets-inline-label-width)")
+        self.button_box = VBox(
+            children=[button for option, button in self.button_from_option.items()],
```
andli28, line 50: Since option isn't used, can we just do `children=[button for button in self.button_from_option.values()],` or make option an _ instead

src/seismometer/controls/selection.py
andli28: Having  `value` be an item in Iterable `value`is confusing.

Something like this would be more concise and clear

```
if value: selected_values = tuple(value)
        else: selected_values = ()
```

andli28: Even better:

`selected_values = tuple(value) if value else ()`

andli28: Less adamant about this one, but you could also combine the subsequent line like

`self.value = selected_values = tuple(value) if value else ()`

gbowlin: FINC

src/seismometer/_api.py
gbowlin: added

src/seismometer/controls/selection.py
```diff
@@ -11,26 +12,25 @@ class SelectionListWidget(ValueWidget, VBox):
...
-            Title to be displayed above the list of buttons.
-        options : tuple[str]
+        options : Iterable[str]
             Selectable options.
-        value : Optional[tuple[str]], optional
+        value : Optional[Iterable[str]], optional
             Subset of options that should be selected by default, by default None.
-        show_title : bool, optional
-            If True, shows the title, by default True.
+        title : Optional[str, optional
```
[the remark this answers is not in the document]
gbowlin: FINC

src/seismometer/controls/selection.py
gbowlin: FINC - going with `children=[button for button in self.button_from_option.values()]`

src/seismometer/controls/selection.py
```diff
@@ -39,57 +39,60 @@ def __init__(self, title: str, options: tuple[str], value: tuple[str] = None, sh
...
-
-    @property
-    def control_value(self) -> tuple:
-        """Value from child widgets."""
-        return tuple(self.value_from_button[x] for x in self.buttons if x.value)
+        self.value = tuple(option for button, option in self.option_from_button.items() if button.value)
 
-    @control_value.setter
-    def control_value(self, new_value: tuple[str]):
+    def _on_value_change(self, change=None):
```
andli28, line 78: Q: parameter `change` isn't used in this function: what is it's purpose

src/seismometer/data/filter.py
```diff
@@ -361,14 +361,42 @@ def none(cls):
...
+
+        rule = cls.all()
+        if not cohort_dict:
+            return rule
+
+        for key in cohort_dict:
+            if not cohort_dict[key]:
+                continue
+            else:
+                rule = rule & cls.isin(key, cohort_dict[key])
```
andli28, line 389:
```
for key, value in cohort_dict.items():
        if value:
            rule = rule & cls.isin(key, value)
```

src/seismometer/controls/selection.py
gbowlin: this is provided by traitlets, it's a dictionary that lets you know what was changed. It has `change["owner"]`, `change["new"]`, and `change["old"]` so you can decided what to do if you want fancy handling.

tests/controls/test_explore.py
```diff
@@ -0,0 +1,267 @@
...
+            nonlocal count
+            count += 1
+
+        widget = undertest.UpdatePlotWidget()
+        widget.on_toggle_code(on_toggle_callback)
+        widget.code_checkbox.value = True
+        assert count == 1
+
+
+class TestExporationBaseClass:
```
andli28, line 54: oops as I was checking off FINC'd comments I see this is misspelled -  TestExporationBaseClass -> Test**Exploration**BaseClass

gbowlin: fixed

andli28 (approved):

diehlbw (approved):

PR #49, opened 2024-07-12
Add an event-specific combine strategy

### LINKED ISSUE #48: Allow configurable strategy for combining scores
## Is your feature request related to a problem? Please describe
Initially performance plots support evaluation at the line-level or at the context-id by taking the maximum model output value.  Maximum windowed value is not the only method for summarization.

## Describe the solution you'd like
We'd like to support configuration to allow min, max, first and last methods.
The core filtering already supports these, so the change is just adding a new config key and then looking to it for the plots.

diehlbw:
# Overview
Closes #48 - Adds a configuration parameter (**combine_strategy**) per event, and makes use of it in model and cohort evaluation plots

## Description of changes
- Adds an optional parameter to the schema for usage_info: combine_strategy.  Allows min, max, first, and last.  Defaults to 'max' (making this backwards compatible for existing config).
- This is the first case where we need to access event-specific config after merge, so updated the access pattern to be dictionary based instead of list.  

## Author Checklist


src/seismometer/configuration/model.py
```diff
@@ -155,6 +155,8 @@ class Event(DerivedFeature):
     """ The value to use if no event is present. """
     usage: Optional[str] = None
     """ The type of event being defined; can be target, intervention, or outcome. """
+    combine_strategy: Optional[Literal["min", "max", "first", "last"]] = "max"
```
diehlbw, line 158: move this to the top

src/seismometer/configuration/config.py
```diff
@@ -326,7 +326,7 @@ def interventions(self) -> dict:
 
         Configured in usage_data as events with usage 'intervention'.
         """
-        return [event.display_name for event in self.events if event.usage == "intervention"]
+        return [event.display_name for event in self.usage.events if event.usage == "intervention"]
```
diehlbw, line 329: make all these dictionaries too

example-notebooks/binary-classifier/usage_config.yml
```diff
@@ -37,5 +37,7 @@ data_usage:
       window_hr: 6
       offset_hr: 0
       usage: target
+      # How to combine multiple *scores* for a context_id when analyzing this event
+      combine_strategy: max
```
gbowlin, line 41: Should we have a comment in the example that max is the default value?

src/seismometer/_api.py
```diff
@@ -176,9 +176,13 @@ def fairness_audit(metric_list: Optional[list[str]] = None, fairness_threshold=1
...
 
-    df = pdh.event_score(sg.data(), sg.entity_keys, score=sg.output, summary_method="max")[
-        [sg.target, sg.output] + sensitive_groups
-    ]
+    df = pdh.event_score(
+        sg.data(),
+        sg.entity_keys,
+        score=sg.output,
+        ref_event=sg.predict_time,
+        summary_method=sg.event_combine_strategy(sg.target),
```
gbowlin, line 184: Since max/min/first/last are aggregate methods and closest to pandas aggregation methods, should we rename these to `aggregation_method`?

More important to me would be can we rename `summary_method` and `sg.event_combine_strategy` as well as the key in config.yml to match?

https://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.core.groupby.DataFrameGroupBy.agg.html

src/seismometer/configuration/config.py
```diff
@@ -286,47 +286,55 @@ def cohorts(self) -> dict:
...
+    def event_group(self, usage_group: str) -> dict[str, Event]:
+        """
+        Returns a dictionary of events indexed by column name and restricted to the specified usage group
+
+        Configured in usage_data as events with usage 'group'.
+        """
+        return {ev.display_name: ev for ev in self.usage.events if ev.usage == usage_group}
 
     @property
     def targets(self) -> list[str]:
```
gbowlin, line 313: Should this be updated now? There are several other places where you added the word Dictionary to the return, but the annotation says list still.

src/seismometer/seismogram.py
```diff
@@ -179,27 +190,27 @@ def target_event(self, event: str):
 
     @property
     def target_cols(self) -> list:
-        return self.config.targets
+        return list(self.config.targets)
```
gbowlin, line 193: Since you are doing a conversion from a dictionary to a list, are these the column names, or event descriptors (the display names)?

This might be worth calling out in a doc string for this property

src/seismometer/seismogram.py
```diff
@@ -179,27 +190,27 @@ def target_event(self, event: str):
...
 
     @property
     def intervention(self):
```
gbowlin, line 196: can we add the return type for this property, it is a string, or a new object?

src/seismometer/seismogram.py
```diff
@@ -179,27 +190,27 @@ def target_event(self, event: str):
...
         """First event in configuration with usage 'intervention'."""
         try:
-            return self.config.interventions[0]
+            return list(self.config.interventions)[0]
         except IndexError as exc:
             raise IndexError("No interventions defined in configuration") from exc
 
     @property
     def outcome(self):
```
gbowlin, line 204: can we add a return type

src/seismometer/seismogram.py
```diff
@@ -179,27 +190,27 @@ def target_event(self, event: str):
...
         """First event in configuration with usage 'outcome'."""
         try:
-            return self.config.outcomes[0]
+            return list(self.config.outcomes)[0]
         except IndexError as exc:
             raise IndexError("No outcomes defined in configuration") from exc
 
     @property
     def comparison_time(self):
```
gbowlin, line 212: can we add a return type

tests/data/loaders/test_load_events.py
```diff
@@ -26,6 +26,15 @@ def __init__(self):
             self.ev_time = "origTime"
             self.ev_value = "origValue"
 
+        # Aid setting a list of events (like that from config) but getting a dict like ConfigProvider returns
+        @property
+        def events(self):
```
gbowlin, line 31: can we add a return type

tests/data/loaders/test_load_events.py
```diff
@@ -26,6 +26,15 @@ def __init__(self):
...
+            return self._events
+
+        @events.setter
+        def events(self, event_list):
```
gbowlin, line 35: can we add a type annotation for event_list?

tests/data/test_pandas_helpers.py
```diff
@@ -180,3 +180,36 @@ def test_string_with_time_stripped_first(self, base):
...
+        ],
+    )
+    def test_three_suffixes_align(self, base, suffix):
+        input = base + suffix
+        assert base == undertest.event_name(input)
+
+    @pytest.mark.parametrize(
+        "input, expected",
+        [
+            (None, None),
```
gbowlin, line 203: Can we add the obvious default test at the top (the bottom two tests are exceptional, even though they also handled the default case)

gbowlin (changes requested):

example-notebooks/binary-classifier/usage_config.yml
diehlbw: Here I was trying to be explicit on what it does since it may be unintuitive; it would get pretty verbose if we add full descriptions for everything.  
The doc string in the config/model specifies the default which also gets revealed in sphinx docs

src/seismometer/_api.py
diehlbw: Done in next commit

src/seismometer/configuration/config.py
diehlbw: good catch

src/seismometer/seismogram.py
diehlbw: Good call. Made it clear these are the latter, the display_name from the event

src/seismometer/seismogram.py
diehlbw: Added here and for outcome, these are the display_names as well.

src/seismometer/seismogram.py
diehlbw: FINC

tests/data/loaders/test_load_events.py
diehlbw: Done

tests/data/test_pandas_helpers.py
diehlbw: Done, if I understood correctly

gbowlin (approved):

PR #57, opened 2024-07-25
Don't use IFrames in notebooks under vscode, add support for altair plot in vscode

### LINKED ISSUE #53: Aequetas wrapper should detect host, since handling of HTML / IFrame is different in jupyter lab, jupyter hub, and vscode.
While this is good for caching, and makes sure that any javascript for the new html segment gets run, it does have two drawbacks:

1. VSCode does not support loading an IFrame source from disk - https://github.com/microsoft/vscode/issues/154722
2. JHub requires some handling for this to propagate to jupyter lab instances - https://stackoverflow.com/questions/60257156/not-able-to-embed-jupyterlab-running-on-jupyterhub-into-an-iframe

_Originally posted by @gbowlin in https://github.com/epic-open-source/seismometer/pull/41#discussion_r1684948314_
            

We can check for the host with something like examining:

```python
import psutil
cmds = [x for  x in psutil.Process().parent().cmdline()]
```

and seeing if `jupyterhub`, `jupyter-lab`, or `vscode-server` appear.

gbowlin:
# Overview

Closes #53  

## Description of changes

Adds the ability to detect the notebook host, so that we can change display behavior based on which host we are in. This is useful when dealing with IFrames in particular, because vscode does not allow them to render local content. 

Since the Altair plot for Aequitas doesn't load well as HTML, we have used the host detection ability to choose whether or not we want to save the plot to HTML/IFrame or just display the plot as is. 

## Testing Instructions

- [ ] make sure you can run the default notebook in vscode and see the fairness control and outcomes
- [ ] make sure you can run the default notebook via jupyter lab and see the fairness control and outcomes

## Author Checklist


src/seismometer/core/nbhost.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @classmethod
+    def get_current_host(cls):
+        cmds = [x for x in psutil.Process().parent().cmdline()]
+
+        if any("jupyter-lab" in cmd for cmd in cmds):
+            return cls.JUPYTER_LAB
+        elif any("jupyterhub" in cmd for cmd in cmds):
+            return cls.JUPYTER_HUB
+        elif any("vscode-server" in cmd for cmd in cmds):
```
diehlbw, line 20: What is actually being matched? And can we more explicit about it (exact match list of pieces)?  

From my test cases, the types of maching seems varied.
 
jupyterhub (jupyterhub-singleuser) is a portion of the default startup command; default seems reused enough, but the partial makes me worry about things like a container or path that happens to have it - made worse by the other comment if we can only be sure about "jupyter".  
vscode-server is just a directory, the worry on this one is more that parents of jupyter-lab might include this.

src/seismometer/core/nbhost.py
```diff
@@ -0,0 +1,22 @@
...
+
+class NotebookHost(Enum):
+    JUPYTER_LAB = 1
+    JUPYTER_HUB = 2
+    VSCODE = 3
+    OTHER = -1
+
+    @classmethod
+    def get_current_host(cls):
+        cmds = [x for x in psutil.Process().parent().cmdline()]
```
diehlbw, line 14: How close could you get from a environment variable approach?  
Looking across at variables with "J" or "VS" (VSCODE_*) seems promising:   
['JPY_SESSION_NAME', 'JPY_PARENT_PID']  

Also, taking inspiration from tqdm autodetect (CLI vs GUI), they do a try around 
`get_ipython = sys.modules['IPython'].get_ipython` for checking ipython setup without importing.


jupyterhub has a lot of options JUPYTERHUB_ variables

src/seismometer/core/nbhost.py
```diff
@@ -0,0 +1,22 @@
...
+
+        if any("jupyter-lab" in cmd for cmd in cmds):
```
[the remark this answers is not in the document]
gbowlin: Moved to ENV variables for now, I was following https://discourse.jupyter.org/t/find-out-if-my-code-runs-inside-a-notebook-or-jupyter-lab/6935/14, which was recommending looking at `psutils`. 

Using `os.environ` look more reliable given your testing of it.

src/seismometer/core/nbhost.py
gbowlin: updated to env vars to fix this.

src/seismometer/core/nbhost.py
gbowlin: FINC

tests/core/test_nbhost.py
```diff
@@ -0,0 +1,34 @@
...
+        "env_key,supports_iframe",
+        [
+            ("VSCODE_CWD", False),
+            ("COLAB_JUPYTER_IP", False),
+            ("JUPYTERHUB_HOST", True),
+            ("JPY_PARENT_PID", True),
+        ],
+    )
+    def test_supports_iframe(self, env_key, supports_iframe):
+        with patch.dict(os.environ, {env_key: "1"}):
```
diehlbw, line 33: I think you need to patch dict on the class to _remove_ the possibilities (or fully replace the os.environ dict here).
I can get these tests to fail by manipulating the shell environment prior to running pytest.

gbowlin: updated

src/seismometer/report/auditing.py
```diff
@@ -82,4 +78,6 @@ def fairness_audit_to_html(
     audit.audit()
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=FutureWarning)
-        audit.summary_plot(metrics=metric_list, fairness_threshold=fairness_threshold).save(path, format="html")
+        altair_plot = audit.summary_plot(metrics=metric_list, fairness_threshold=fairness_threshold)
+
+    return altair_plot
```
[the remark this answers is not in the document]
gbowlin: updated

src/seismometer/_api.py
```diff
@@ -203,7 +204,7 @@ def generate_fairness_audit(
     per_context: bool = False,
     metric_list: Optional[list[str]] = None,
     fairness_threshold: float = 1.25,
-) -> HTML | IFrame:
+) -> HTML | IFrame | Any:
```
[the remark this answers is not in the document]
gbowlin: I didn't want to do the try import or the string matching when all that really matters is that its displayable in an `ipywidgets.Output` widget.  Leaving as is for now.

diehlbw (dismissed):

diehlbw (approved):

diehlbw (approved):

PR #72, opened 2024-08-07
Fix and simplify window merge

### LINKED ISSUE #71: Merge without contextid can create superfluous columns
## Problem Summary
When creating cohorts that do NOT use contextid (but including it in the events data), subsequent cohorts can hit errors around creation of duplicate rows:

FutureWarning: Passing 'suffixes' which cause duplicate columns {'ctxid_x'} in the result is deprecated and will raise a MergeError in a future version.

## Impact
This can be a hardstop in later versions, requiring the management of fake ctx ids.
Could also be hit if events had other extraneous columns.

## Possible Solution
When merging, conditionally reduce the event frame to the needed columns.

diehlbw:
# Overview

Closes #71 - prevent context_id_x columns.  
Simplifies merge to be a clear 'first' event within window.
  

## Description of changes
- reduce event frame to known columns during merge steps
- Remove coalesce logic for merging on late predictions and a subset of missing times  
- Add separate handling for merging events that never have times
- Add logging back when those missing cases are being hit and what the behavior will be

Needed test updates show fairly nicely that the main change is -1+Time pairs are now 0+NaT, and NaT times (early-out of window) also drop label.  
Minimal impact, since most accesses of the event frames filter to > -1, which is biased toward binary classifiers anyway. Did NOT remove that filtering (yet?).  
This simplification is a good step toward supporting more strategies, which a new collaborator has asked about doing.

## Author Checklist

### DIFF
--- a/changelog/71.bugfix.rst
+++ b/changelog/71.bugfix.rst
@@ -1,0 +1,3 @@
+Handle merging without context id; assumes events dataframe columns.
+Remove coalecsing logic from merging, now assumes an event type always has times (or never does). Preps for supporting strategies.
+Hardens restriction of events (with times) to occur after prediction time + window offset.
--- a/changelog/71.bugfix.rst
+++ b/changelog/71.bugfix.rst
@@ -1,3 +1,1 @@
 Handle merging without context id; assumes events dataframe columns.
-Remove coalecsing logic from merging, now assumes an event type always has times (or never does). Preps for supporting strategies.
-Hardens restriction of events (with times) to occur after prediction time + window offset.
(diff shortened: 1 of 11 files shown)


firste:
Have you had a chance to test the performance of this change? Seems like it should be neutral or maybe even a bit positive.

firste:
Can you also add a note to /docs/getting_started/index.rst data gathering section about the stronger assumptions related to times for events?

firste (changes requested):

diehlbw:
> Have you had a chance to test the performance of this change? Seems like it should be neutral or maybe even a bit positive.

Nothing super rigorous, I timed the run_startup on our dataset of 15M predictions (3 cohorts, 3 
events one of which is windowing):
v0.1.1 = 4min 36s wall (4min03s CPU)
v0.2.0 = 4min 36s wall (4min02s CPU)
new    = 2min 46s wall (2min26s CPU)

As guessed, doing a couple fewer merges is looking like a nice improvement on setup.

diehlbw:
> Can you also add a note to /docs/getting_started/index.rst data gathering section about the stronger assumptions related to times for events?

I made a small change there but it didn't seem that relevant considering existing content.  I updated the user_guide as well and noted that the datamodel already describes it as what we do after this update

firste (approved):

firste:
This is good to go from my perspective

PR #81, opened 2024-08-23
Limit display widths for api components, allow row wrapping of cohort selections. 

gbowlin:
# Overview

Closes #xxx 

## Description of changes
* Limits control max-widths to 1200px in most cases, forcing a row wrap when needed (Sphinx allows up to 1300px in content width right now, which includes some margin for output numbers, etc)
image
image

*fixes `sm.cohort_list` aggregation for multiple targets 
image

* update the filtering/error message for the fairness sample at high thresholds (where there are no positive samples)
image

## How to test
You can use the classifier bin notebook to pull down a sample dataset, and then edit the `usage_config.yml` to add additional cohorts like this

```yml

  # This list defines available cohort options for the default selectors
  cohorts:
    - source: age
      display_name: Age
    - source: race
      display_name: Race
    - source: gender
      display_name: Gender
    - source: gender
      display_name: Gender2
    - source: gender
      display_name: Gender3
    - source: gender
      display_name: Gender4
```

## Author Checklist


src/seismometer/_api.py
```diff
@@ -348,13 +348,27 @@ def _cohort_list_details(cohort_dict: dict[str, tuple[Any]]) -> HTML:
     groups = data[cfg.entity_keys + cfg.output_list + intervention_cols + outcome_cols + target_cols].groupby(
         target_cols
     )
-    aggregation = {cfg.entity_id: ["count", "nunique"]}
+    columns = []
+    for c in ["count", "nunique"]:
```
diehlbw, line 352: This loop is a bit unexpected,  pd.Groupby.agg() can usually handle lists of agg functions.
Can you look into passing a dict in, and then overwriting column names?

The gist:
```python
stat_dict = {k: ['mean'] for k in list(data.select_dtypes(include=float))}
stat_dict.update({cfg.entity_id: ['nunique', 'count'], 
                  cfg.context_id: ['nunique'], ...})
groupstats = groups.agg(stat_dict)
groupstats.columns = [' '.join(c) for c in groupstats.columns]
```

src/seismometer/controls/styles.py
```diff
@@ -3,9 +3,10 @@
 WIDE_LABEL_STYLE = {"description_width": "120px"}
 
 BOX_GRID_LAYOUT = Layout(
-    align_items="flex-start", grid_gap="20px", width="100%", min_width="300px", max_width="1400px"
+    align_items="flex-start", grid_gap="20px", width="100%", min_width="300px", max_width="1200px"
```
diehlbw, line 6: Curious - why does this one move FROM 1400px to 1200px and everywhere else seems to move TO using 1400px?  
some kind of 200px buffer?

src/seismometer/html/resources/info.html
```diff
@@ -3,32 +3,33 @@
         text-align: left !important;
     }
 </style>
-
-<h3>Summary</h3>
-The preloaded data covers {{ num_predictions }} predictions over {{ num_entities }} entities from the dates {{ start_date }} to {{ end_date }}.
-<table>
-    <thead>
+<div style="width: 100%; max-width: 1400px;">
```
diehlbw, line 6: Is there a way to parametrize this `"width: 100%; max-width: 1400px;"`? or make it some kind of constant?

If it's more burdensome/obfuscating than its worth then it is fine, but have to ask

tests/resources/rendered_templates/info.html
```diff
@@ -3,28 +3,29 @@
...
-<h4>Plot Functions</h4>
-<ul>
-    <li><code>sm.model_evaluation()</code> - Overall performance across thresholds</li>
-    <li><code>sm.cohort_evaluation(cohort_group)</code> - Performance split by specified cohort</li>
-    <li><code>sm.plot_outcome(outcome, intervention, cohort)</code> - Compare trends of interventions to outcomes</li>
-</ul> 
+    </table>
+    <h4>Plot Functions</h4>
+    <ul>
+        <li><code>sm.model_evaluation()</code> - Overall performance across thresholds</li>
```
diehlbw, line 27: Not your PR, but are these even what we want to call out anymore? or should the refs be replaced ExploreXXX() classes?

diehlbw (changes requested):

src/seismometer/controls/styles.py
gbowlin: moved everything to 1200px due to sphinx, this was an accidental exception.

tests/resources/rendered_templates/info.html
gbowlin: yeah, just fixing this to 

``` html
        <li><code>sm.ExploreModelEvaluation()</code> - Overall performance across thresholds</li>
        <li><code>sm.ExploreCohortEvaluation()</code> - Performance split by specified cohort</li>
        <li><code>sm.ExploreCohortOutcomeInterventionTimes()</code> - Compare trends of interventions to outcomes</li>
```

src/seismometer/html/resources/info.html
gbowlin: We could pass the style in as a parameter, and then hard code it on the load template function. that's not too bad, and unit tests would catch it by default. 

```html
<div style="{{ display_style }}">
```

The second option is to display an HTML element earlier that includes a CSS class in a style tag so that we can just reference that style. This would be ideal, but I'm not sure what the best hook is to return the css since we need that to be displayed in the notebook to get the CSS, generating the HTML element is not enough. 

```python
def _set_ux_css()
    from IPython.display import HTML
    return HTML(Path('custom.css').read_text())
```

The second option would be better, since it gives us a path to add more CSS content in the future, but since I can't think of a good way to load an extra CSS file (other than requiring themes/extensions) I'm going to stick with the first option.

src/seismometer/_api.py
gbowlin: I was running into an crash based on 
image

So the kludge was to deal with not being able to pass a dictionary. 
I looked into the core code and applied the right fix (thanks for this comment)

src/seismometer/_api.py
```diff
@@ -335,26 +335,35 @@ def _cohort_list_details(cohort_dict: dict[str, tuple[Any]]) -> HTML:
...
+    stat_dict = {k: ["mean"] for k in float_cols}
+    stat_dict.update({cfg.entity_id: ["nunique", "count"], cfg.context_id: ["nunique"]})
+
+    groupstats = groups.agg(stat_dict)
+    groupstats.columns = [pdh.event_name(x) for x in float_cols] + [
+        f"Unique {cfg.entity_id}",
+        f"{cfg.entity_id} Count",
+        f"Unique {cfg.context_id}",
+    ]
+    groupstats.index.rename([pdh.event_name(x) for x in target_cols], inplace=True)
```
diehlbw, line 363: This line is still creating issues.
MultiIndex (multiple defined targets) seems to work, but it looks like it fails if only one target is defined.  I think rename on single index can't handle being passed a list (my suspicion is its due to pd.Index.name vs pd.MultiIndex.names)

diehlbw (changes requested):

src/seismometer/_api.py
gbowlin: fixed, when we move this to use `great_tables`, or move this out of `_api` we should make adding unit tests around it a requirement.

diehlbw (approved):

PR #100, opened 2024-10-10
Remove explicit 'invalid value' imputation

diehlbw:
# Overview
Several reports of confusion around -1 imputation for "invalid" predictions.
It only made complete sense in loose typed binary classifications, so was going to start causing issues with stricter datatypes as more variation is used.

## Description of changes
Remove the actual -1 imputation (based on time comparisons)  
Update the validation check to directly make the time comparison, specifically  
  Changed pdh.valid_event -> filtered frame to pdh.is_valid_event -> mask
Use pdh.is_valid_event more comprehensively instead of oneoff checks

## Author Checklist


src/seismometer/data/pandas_helpers.py
```diff
@@ -543,9 +535,11 @@ def event_value_name(event_value: str) -> str:
     return val
 
 
-def valid_event(dataframe: pd.DataFrame, event: str) -> pd.DataFrame:
-    """Filters a dataframe to valid predictions, where the event value has not set to -1."""
-    return dataframe[dataframe[event_value(event)] >= 0]
+def is_valid_event(dataframe: pd.DataFrame, event: str, ref: str) -> pd.DataFrame:
```
gbowlin, line 538: Should is_valid_event include the event's merge_strategy?

src/seismometer/data/pandas_helpers.py
```diff
@@ -543,9 +535,11 @@ def event_value_name(event_value: str) -> str:
...
+    """Creates a mask if events occur after the reference time, defaults to True."""
```
gbowlin, line 539: If the event is not in the dataframe, shouldn't this all be false so we end up with no rows?

src/seismometer/data/timeseries.py
```diff
@@ -77,9 +79,9 @@ def _limit_data(
 ) -> pd.DataFrame:
     """Reduces the data to only include valid data points."""
     include = dataframe[reftime].notna()
-    # boolean events - filter out -1 where invalid
+
     if boolean_event:
-        include = include & (dataframe[event_col] >= 0)
+        include = include & (is_valid_event(dataframe, event_col, reftime))
```
gbowlin, line 84: Just calling out this is the important change, where knowing the event + ref time we can know that we predicted the correct value.

src/seismometer/data/pandas_helpers.py
gbowlin: I am curious if we ever want a model to predict events in the past? - For example, we may want to check if an event in the past was erroneously added to the chart (say a DX code was added, but was incorrect, or the Observation status assigned to a patient was incorrect)

diehlbw: Currently we don't have any validation/restriction on merge_strategy so isnt needed (yet).

In your scenario, I don't know think that predicttime is relevant.  And this kind of supports my implementation of "default True" keeping everything.
Unless it's predicting the event+time, I think we'd want to make sure we could strip times since they aren't relevant.

src/seismometer/data/pandas_helpers.py
diehlbw: I can update doc if you think it'd be useful, but have generally been thinking about timing being  respected as a filter only when it exists.  If it's "unknown", I wouldn't want this to hide it.

I could also update the name to is_not_invalid_event, if you think the default is more intuitive that way.

gbowlin: I think just updating the header to something of the effect : "If the event has a timestamp, verify the timestamp is after the prediction"

gbowlin (approved):

PR #111, opened 2024-12-02
add example data loader function

### LINKED ISSUE #105: Example notebooks should put data loading into a single function call.
_Originally posted by @diehlbw in https://github.com/epic-open-source/seismometer/pull/99#discussion_r1812702070_

Basically given a url, download all the requisite files based on our general expectations.

firste:
# Overview

Closes #105 

## Description of changes
adds a download_example_dataset function to Seismometer singleton object. There are a handful of implementation decisions worth calling out. 
1. internally is has an allow list for which datasets are supported. 
2. Since Seismometer.run_startup requires the data to be present before being run, any side effects that it has like setting up the singleton object, loading configuration, and initializing loggers, cannot be depended upon.

## Author Checklist


src/seismometer/__init__.py
```diff
@@ -72,3 +72,46 @@ def run_startup(
     # Surface api into namespace
     s_module = importlib.import_module("seismometer._api")
     globals().update(vars(s_module))
+
+
+def download_example_dataset(dataset_name: str, branch_name: str = "main"):
+    """
+    Downloads an example dataset from the specified branch to local data/ directory.
+
+    Args:
```
diehlbw, line 81: Can you change the docstring format to numpy-style?

Instead of Args it's Parameters like:
```python
    Parameters
    ----------
```
Bigger thing is to add this function to one of the .rst under docs/reference/ and make sure it shows up consistently

diehlbw (changes requested):

src/seismometer/__init__.py
firste: fixed

example-notebooks/binary-classifier/events.parquet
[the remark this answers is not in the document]
firste: cleaned this up

Issue #51, opened 2024-07-18
Enhance usage config to include score bounds/type


gbowlin:
## Is your feature request related to a problem? Please describe

When describing model outcomes, we expect a threshold to come from the user, but its not always clear what type of outcome we have (0-100 or 0.00 to 1.00 for percentiles versus probabilities)

## Describe the solution you'd like

For clarity, output configuration should be more than just a column name, but should include expected valid values that Seismometer can filter based on, or provide as limits. Similar to how we use a custom class for the `cohort` columns. 

For example:

```
primary_output: Primary Model Score
outputs:
  - source: ScoreName1
    display_name: Primary Model Score
    type: numeric
    lower_bound: 0
    upper_bound: 1
  - source: ScoreName2
    display_name: A Second Score
    type: numeric
    lower_bound: 0
    upper_bound: 100
  - source: ScoreName3
    display_name: A Third Score for completeness
    type: numeric
    lower_bound: -12
    upper_bound: 144
```

Where `type` could potentially be extended to allow categorical scores in the future.

gbowlin:
For bounds I would assume we follow the numpy standard of include lower bound exclude upper bound, and if we need to we can come back with an option to select how you want bounds to work.  

For clip vs clear (set to `nan`), this should probably an option with clip being the default.

diehlbw:
May break this out into a separate issue, but want to track that annotations can be improved alongside discretization of score bounds.  

We currently have a couple methods in lines to add labels to our plots, with *mostly* baked in assumptions around being 0-1.  The mostly has caused around scaling.  
Once we handle scaling more consistently closer to data-loading, functions/plots can have stronger opinions - the annotations would likely assume percentile (and label with percentage).

PR #125, opened 2024-12-31
Scaling roc

### LINKED ISSUE #124: ROC plot issue
ROC in the 2x3 plot scales incorrectly

diehlbw:
# Overview

Closes #124 

## Description of changes
Changed model evaluation code (binary classifiers) to scale the threshold info for ROC annotations

image

### Testing
Verify the docs include a notebook with corrected annotations like that above - verify all 6 plots  
Alter the score data to be from 0-100,  sm.Seismogram().dataframe['Risk30DayReadmission']*100 rerun and verify it is still ok.

## Author Checklist

### DIFF
--- a/src/seismometer/data/performance.py
+++ b/src/seismometer/data/performance.py
@@ -390,7 +390,9 @@ def calculate_nnt(arr: np.ndarray, rho: Optional[Number | None] = None) -> np.nd
 
 
 @export
-def calculate_eval_ci(stats: pd.DataFrame, truth: pd.Series, output: pd.Series, conf: Number = 0.95) -> dict:
+def calculate_eval_ci(
+    stats: pd.DataFrame, truth: pd.Series, output: pd.Series, conf: Number = 0.95, force_percentages=False
+) -> dict:
     """
     Calculate confidence intervals for ROC, PR, and other performance metrics from a stats frame.
 
@@ -404,6 +406,8 @@ def calculate_eval_ci(stats: pd.DataFrame, truth: pd.Series, output: pd.Series, 
         The series of model output associated with the stats frame.
     conf : Number, optional
         The confidence level for calculation, by default 0.95.
+    force_percentages : bool, optional
+        Flag to indicate that outputs should be converted to percentages (0-100), by default False.
 
     Returns
     -------
@@ -416,6 +420,9 @@ def calculate_eval_ci(stats: pd.DataFrame, truth: pd.Series, output: pd.Series, 
     conf = confidence_dict(conf)
     if truth is None or output is None:
         return conf, None, None, None
+
+    if force_percentages:
+        output = as_percentages(output)
 
     roc_conf = ROCConfidenceParam(conf)
     aucpr_conf = PRConfidenceParam(conf)
--- a/src/seismometer/plot/mpl/_lines.py
+++ b/src/seismometer/plot/mpl/_lines.py
@@ -251,14 +251,13 @@ def add_radial_score_thresholds(
     y : np.ndarray
         Array of y-values.
     labels : list[str]
-        Array of thresholds to display.
-    colorIx : int, optional
-        Index of line being annotated in order to match color, by default 0.
-    n_scores : int, optional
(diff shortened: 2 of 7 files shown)


src/seismometer/plot/mpl/binary_classifier.py
```diff
@@ -61,10 +61,15 @@ def evaluation(
         prevalence = prevalence / stats.loc[0, ["TP", "FP", "TN", "FN"]].sum()
 
     roc_data = ci_data["roc"]
+    roc_thresholds = roc_data["Threshold"]
```
gbowlin, line 64: Can you add a comment to ci_data to include something about the ROC data? It feels like that should be fixed to a normalized format, sicne Threshold could be percentage or precentile.

src/seismometer/plot/mpl/binary_classifier.py
```diff
@@ -61,10 +61,15 @@ def evaluation(
...
+    # Normalize roc_data like stats; ignore infs inserted at beginning
+    if not (roc_thresholds[2:] > 2).any():
```
gbowlin, line 66:
```suggestion
    if not (roc_thresholds[2:] > 1).any():
```

1.5 is also greater than 1, at the very least `>=2` to make sure we catch that edge case.

gbowlin (changes requested):
Can we move the implicit formatting check out, and do a discrete check instead?

src/seismometer/plot/mpl/binary_classifier.py
diehlbw: Updated.  This method was due to _old_ sklearn behavior, and the slicing skips that insertion regardless of value.

src/seismometer/plot/mpl/binary_classifier.py
diehlbw: Moved the logic into the ci calculation, and reused the as_percentages method that binary_stats does

src/seismometer/plot/mpl/binary_classifier.py
diehlbw: And then deleted...

src/seismometer/plot/mpl/_lines.py
```diff
@@ -251,14 +251,13 @@ def add_radial_score_thresholds(
...
-    n_scores : int, optional
-        Integer number of scores to display, if length of arrays is greater, will automatically
-        select 10 values to shows based on an assumed range of labels being 0-1 (default: 10).
+        Array of labels to display.
+    thresholds : list[Number]
+        List of threshold points to label.
     Q : int, optional
         The quadrant 1-4 to assume for offsets, by default 1.
+    colorIx : int, optional
+        Index of line being annotated in order to match color, by default 0.
```
gbowlin, line 260: I think you can now remove

gbowlin: image

diehlbw: Annoyingly... this is not the case.  
This instance isn't about scores nor roc_info, it's around the thresholds being labeled.  In our examples this is the 0.2 and 0.8 that are displayed as 20 and 80.

Added a comment https://github.com/epic-open-source/seismometer/issues/51#issuecomment-2568417887 to improve this alongside making score bounds more discrete, which I assume will include scaling on load.
Also made these functions private to limit usage with the possibility that they will change dramatically alongside that enhancement.

src/seismometer/data/performance.py
```diff
@@ -404,6 +406,8 @@ def calculate_eval_ci(stats: pd.DataFrame, truth: pd.Series, output: pd.Series,
         The series of model output associated with the stats frame.
     conf : Number, optional
         The confidence level for calculation, by default 0.95.
+    force_percentages : bool, optional
```
gbowlin, line 409: Should we add a unit test for this use case?

gbowlin (commented):
Looks fine, can we add a unit test around forcing precents?

gbowlin (approved):

PR #134, opened 2025-02-18
Support reading parquet files from non-Pandas sources

ckchow:
# Overview
Creating a parquet file with e.g. Polars will write a file without pandas-related metadata.  Pandas is able to read files without this metadata, but a possible bug in the API requires that calls to `pandas.read_parquet` for these kinds of files be made such that the `columns` parameter must *strictly* be a list of strings and not a set. If Pandas metadata is present in the parquet file, Pandas will accept a set of columns and normalize it to a list internally. If the metadata is not present, calling the function with a set will yield a stacktrace like the following:

```
/home/seismo/workspace/tests/data/loaders/test_load_predictions.py:124: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/seismo/workspace/src/seismometer/data/loader/prediction.py:43: in parquet_loader
    dataframe = pd.read_parquet(config.prediction_path, columns=actual_columns)
/home/seismo/.local/lib/python3.11/site-packages/pandas/io/parquet.py:667: in read_parquet
    return impl.read(
/home/seismo/.local/lib/python3.11/site-packages/pandas/io/parquet.py:274: in read
    pa_table = self.api.parquet.read_table(
/home/seismo/.local/lib/python3.11/site-packages/pyarrow/parquet/core.py:1843: in read_table
    return dataset.read(columns=columns, use_threads=use_threads,
/home/seismo/.local/lib/python3.11/site-packages/pyarrow/parquet/core.py:1485: in read
    table = self._dataset.to_table(
pyarrow/_dataset.pyx:553: in pyarrow._dataset.Dataset.to_table
    ???
pyarrow/_dataset.pyx:399: in pyarrow._dataset.Dataset.scanner
    ???
pyarrow/_dataset.pyx:3594: in pyarrow._dataset.Scanner.from_dataset
    ???
pyarrow/_dataset.pyx:3512: in pyarrow._dataset.Scanner._make_scan_options
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   ValueError: Expected a list or a dict for 'columns', got <class 'set'> instead.
```

## Description of changes
To prevent this from happening, just explicitly cast the columns variable in `parquet_loader` to a list of strings after doing the set intersection. This shouldn't be a significant source of overhead because as far as I can tell this is only done once per file load.

## Author Checklist

### DIFF
--- a/tests/data/loaders/test_load_predictions.py
+++ b/tests/data/loaders/test_load_predictions.py
@@ -5,6 +5,8 @@ from unittest.mock import Mock
 import numpy as np
 import pandas as pd
 import pandas.testing as pdt
+import pyarrow as pa
+import pyarrow.parquet as pq
 import pytest
 from conftest import tmp_as_current  # noqa
 
@@ -70,11 +72,29 @@ def parquet_setup():
     return fake_config(file)
 
 
+def non_pandas_parquet_setup():
+    file = Path("predictions.parquet")
+
+    data = pred_frame()
+    # remove the pandas metadata in the parquet file to simulate the data
+    # coming from a different parquet source (e.g. polars)
+    table = pa.Table.from_pandas(data).replace_schema_metadata({})
+
+    pq.write_table(table, file)
+    return fake_config(file)
+
+
 # endregion
 
 
 # region Tests
-@pytest.mark.parametrize("setup_fn,load_fn", [[parquet_setup, undertest.parquet_loader]])
+@pytest.mark.parametrize(
+    "setup_fn,load_fn",
+    [
+        [parquet_setup, undertest.parquet_loader],
+        [non_pandas_parquet_setup, undertest.parquet_loader],
+    ],
+)
 @pytest.mark.usefixtures("tmp_as_current")
 class TestPredictionLoad:
     def test_load_all_columns(self, setup_fn, load_fn):
@@ -107,7 +127,10 @@ class TestPredictionLoad:
 
     @pytest.mark.parametrize(
         "extra_columns",
-        [pytest.param([], id="all features"), pytest.param(["column1", "column2", "column3"], id="full features")],
+        [
+            pytest.param([], id="all features"),
+            pytest.param(["column1", "column2", "column3"], id="full features"),
+        ],
     )
     def test_target_inclusion_is_renamed(self, extra_columns, setup_fn, load_fn):
         config = setup_fn()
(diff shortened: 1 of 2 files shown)


gbowlin (commented):
Thanks for the submission and the fix to support non-pandas based Parquet files!

I left a minor cleanup task/nitpick, not even related to your code, but relevant now that we are testing two different sources of parquet files. 

Also thank you for not adding a polars as a testing dependency and using pyarrow instead.

tests/data/loaders/test_load_predictions.py
```diff
@@ -70,11 +72,29 @@ def parquet_setup():
...
 # endregion
 
 
 # region Tests
-@pytest.mark.parametrize("setup_fn,load_fn", [[parquet_setup, undertest.parquet_loader]])
+@pytest.mark.parametrize(
+    "setup_fn,load_fn",
+    [
+        [parquet_setup, undertest.parquet_loader],
+        [non_pandas_parquet_setup, undertest.parquet_loader],
```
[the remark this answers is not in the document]
ckchow: Accepted your suggestion, thanks for the review.

tests/data/loaders/test_load_predictions.py
```diff
@@ -70,11 +72,29 @@ def parquet_setup():
...
+
 # endregion
 
 
 # region Tests
-@pytest.mark.parametrize("setup_fn,load_fn", [[parquet_setup, undertest.parquet_loader]])
+@pytest.mark.parametrize(
+    "setup_fn,load_fn",
+    [
+        [pandas_parquet_setup, undertest.parquet_loader],
```
gbowlin, line 94: You also need to update line 66, and anything that references it so that it has the new name.

gbowlin (changes requested):

tests/data/loaders/test_load_predictions.py
ckchow: Updated ref in test_load_predictions, tests pass. I found another structurally similar test in test_load_events and copied the non-pandas-parquet function over just for completeness.

gbowlin (approved):

PR #138, opened 2025-03-19
Different number of positive cases in cohort summary and model evaluation plot

### LINKED ISSUE #137: Different number of positive cases in cohort summary and model evaluation plot
## Problem Summary
When aggregating data per context, in `show_cohort_summaries` (the earliest score) and `plot_model_evaluation` (based on the merge strategy, by default, max score) we use different strategies. This results in discrepancy on the number of `True`, `False` cases.

For example: (Note that times are relative to a fixed time, event_time is always after score_time, the other case is replaced with NaT)
```
        entity     prediction           score    score_time   event_time  event_value
0          1              0             0.20       +1hr         +2hr            1
1          1              1             0.30       +2hr         +2hr            1
2          1              2             0.50       +3hr          NaT            0
3          1              3             0.01       +4hr          NaT            0
```
Since score_time is ordered, in `show_cohort_summaries`, we pick the first row, that is the earliest score_time for the encounter
  
Image

But in plot_model_evaluation, we pick the row with the max score which in this case is the third row with event_value of 0.

Image

Image

 

## Impact
The number of `True/False` cases in the target do not match in the mentioned visuals.

## Suggested fix
Update how per context (i.e., per entity) data is aggregated to:
1. Make it consistent throughout,
2. Only consider scores in the time window when applying the merge strategy.

MahmoodEtedadi:
# Overview

Closes #137, #144

## Description of changes

We update how per-context calculations are done in `pdh.event_score` and make it consistent between different visualizations/table summaries.

### Functional Testing
1. Run the example notebook and make sure that all visualizations work with or without per context checkbox.
2. Run `sm.show_cohort_summaries` with `by_target=True, by_score=True` and make sure the breakdown of entities by score, target adds up to the total number of entities given.
image

image
3. Make sure the number of positive/negative predictions in `sm.show_cohort_summaries` is the same as the number of positive/negative cases in `sm.ExploreModelEvaluation` (Predicted Probability graph).
image
Check the same for the number of positive/negative entities in `sm.show_cohort_summaries` and the number of positive/negative cases in `sm.ExploreModelEvaluation` with per context checkbox selected.

4. Make sure #144 is fixed:
i. Run the first code block in the example notebook to download the config/data files.
ii. In `usage_config`, remove `context_id`.
iii. Comment out `sm.download_example_dataset('diabetes-v2')` from the first code block.
iv. Run the example notebook and make sure there are no errors. 

### Performance Testing
1. Run the example notebook from this branch and the main branch and make sure notebook runs as fast as (maybe a bit slower in `sm.show_cohort_summaries`, we had to make it more computationally expensive for the sake of consistent number of positive cases).
2. With per context checkbox selected, visualization should run faster (in some cases significantly faster) than before. 
3. You can repeat 1, 2 with a larger dataset (e.g., use `repeated_df = pd.concat([df] * 10, ignore_index=True)` to multiply the number of rows in dataset by 10; then replace `predictions.parquet` with the new dataframe).

  
## Author Checklist


src/seismometer/data/pandas_helpers.py
```diff
@@ -440,44 +443,41 @@ def event_score(
...
 
-    elif aggregation_method == "min":
-        ref_col = ref_score
+    if len(df.index) == 0:
+        return
+
+    pks = [c for c in pks if c in df.columns]
+
+    if aggregation_method == "max":
+        df = df.sort_values(by=[event_value, ref_score], ascending=True)
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

src/seismometer/data/pandas_helpers.py
```diff
@@ -440,44 +443,41 @@ def event_score(
...
-            return gf.idxmin()
+    elif aggregation_method == "min":
+        df = df.sort_values(by=[event_value, ref_score])
+        ix = df.drop_duplicates(subset=pks).index
 
     # merged frame has time columns only for events in appropriate time window,
     # implicitly reduces to positive label (need method to re-add negative samples)
     elif aggregation_method == "last":
+        df = merged_frame[merged_frame[event_time].notna()]
+        df = df.sort_values(by=event_time, ascending=True)
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

tests/data/test_merge_window.py
```diff
@@ -888,8 +888,10 @@ def test_merge_first(self):
             [1] * 10, [x for x in range(1, 11)], event_name, event_times=event_times, event_values=[1] * 10
         )
 
-        expected_vals = [1, 0]
-        expected_times = ["2024-01-01 00:00:01", pd.NA]
+        expected_vals = [0, 0]
+        # For the first row, eventhough "2024-01-01 01:00:00" is a valid time, we first merge with "first" strategy
+        # which gives us "2024-01-01 00:00:01" and then filter it out as it is before the scoring time.
+        expected_times = [pd.NA, pd.NA]
```
gbowlin, line 894: Should we have something where the expected value is 1 still?

tests/data/test_merge_window.py
```diff
@@ -935,8 +937,8 @@ def test_merge_nearest(self):
             [1] * 5 + [2] * 5, [x for x in range(1, 11)], event_name, event_times=event_times, event_values=[1] * 10
         )
 
-        expected_vals = [1, 1]
-        expected_times = ["2024-01-01 11:00:00", "2024-01-01 11:00:00"]
+        expected_vals = [0, 1]
+        expected_times = [pd.NA, "2024-01-01 11:00:00"]
```
gbowlin, line 941: I think this test was setup so that we had a nearest before and a nearest after, it might be good to make these cases more explicit (maybe using params?) 

How does your windowing affect nearest, do we need to update the leadtime_hours to be negative so that we can get an event in the future?

tests/data/test_events.py
```diff
@@ -35,6 +35,9 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
             & (expected_frame["ref_event"] == ref_event),
             aggregation_method,
         ]
-
-        actual = undertest.event_score(input_frame, ["Id", "CtxId"], "ModelScore", ref_event, aggregation_method)
-        assert actual["ModelScore"].tolist() == expected_score.tolist()
+        if ref_event in ["PredictTime", "Reference_5_15_Time"] and aggregation_method in ["min", "max"]:
```
gbowlin, line 38: Can we create a new unit test expressly for the issue you are fixing, so that we have a direct test we can link to as "this would have failed for the prior code, but now passes"?

gbowlin (changes requested):
Can we improve the unit tests for your example. Mostly I want to explicitly call out how max/min now work with the windowing.

src/seismometer/data/pandas_helpers.py
```diff
@@ -147,9 +147,12 @@ def merge_windowed_event(
...
     if window_hrs is not None:  # Clear out events outside window
-        max_lookback = pd.Timedelta(window_hrs, unit="hr") + min_offset  # keep window the specified size
-        filter_map = predictions[predtime_col] < (predictions[r_ref] - max_lookback)
+        max_lookback = pd.Timedelta(window_hrs, unit="hr")  # r_ref has already been moved by min_offset.
+        filter_map = (predictions[predtime_col] < (predictions[r_ref] - max_lookback)) | (
+            predictions[predtime_col] > (predictions[r_ref])
+        )
         predictions.loc[filter_map, [event_val_col, event_time_col]] = pd.NA
 
     predictions = post_process_event(
```
diehlbw, line 158: Not your MR - can you fix the spelling issues in post_process_event?  
"analyes" & "imutations" for ctrl+f

src/seismometer/data/pandas_helpers.py
```diff
@@ -440,44 +443,41 @@ def event_score(
...
 
-    ref_score = _resolve_score_col(merged_frame, score)
-    if aggregation_method == "max":
-        ref_col = ref_score
+    # ref_event = ref_event or sg.target
+    event_time = _resolve_time_col(merged_frame, ref_event)
 
-        def apply_fn(gf):
-            return gf.idxmax()
+    if aggregation_method in ["first", "last"]:
```
diehlbw, line 449: Codestyle: 
Are we worried that aggregation methods will get this bit wrong (again)?
Or are we confident all aggregation methods exist? My comment becomes moot if you are event reasonably confident.

Otherwise, this moves away from the organization I'd prefer (opinion).  
Acknowledging potential for duplication, rather than having two if-else blocks on aggregation_method (this new one and the older one)...
I think it'd be more consistent/modular if you combine the steps of logic and maybe even move it to a case match - encapsulate the functionality for each method in a single place.

src/seismometer/data/pandas_helpers.py
```diff
@@ -440,44 +443,41 @@ def event_score(
...
-    elif aggregation_method == "min":
-        ref_col = ref_score
+    if len(df.index) == 0:
+        return
+
+    pks = [c for c in pks if c in df.columns]
+
+    if aggregation_method == "max":
+        df = df.sort_values(by=[event_val, ref_score], ascending=False)
+        ix = df.drop_duplicates(subset=pks).index
```
diehlbw, line 463: If my other code reorg seems reasonable, I think we could embrace it further.  
Rather than have these cases focus on determining the ix that is then applied, just produce the dataframe.

src/seismometer/data/pandas_helpers.py
```diff
@@ -440,44 +443,41 @@ def event_score(
...
 
-        def apply_fn(gf):
-            return gf.idxmin()
+    elif aggregation_method == "min":
+        df = df.sort_values(by=[event_val, ref_score])
```
diehlbw, line 466: Curious - what kind of performance gains were you seeing for the sort + dropduplicates over a groupby (+ sort?) + first  

Mostly wondering because in my mental model the sort after groupbys would likely be pretty trivial (I assume very few label/event_val changes for a contextid - which might be too limiting).

src/seismometer/data/pandas_helpers.py
```diff
@@ -440,44 +443,41 @@ def event_score(
...
+        ix = df.drop_duplicates(subset=pks).index
 
     # merged frame has time columns only for events in appropriate time window,
     # implicitly reduces to positive label (need method to re-add negative samples)
     elif aggregation_method == "last":
+        df = merged_frame[merged_frame[event_time].notna()]
```
diehlbw, line 472: This gets at my code organization comment  a little - but is this effectively a noop after line~450?

src/seismometer/data/pandas_helpers.py
diehlbw, line 463: I'm sure I'll understand once I start doing some functional tests but my read:  
pks is ALL columns (ln 459)  -> drop_duplicates on pks removes only identical rows.

Naively I'd think we'd want to drop on the subset [pks - ref_score]

diehlbw (changes requested):

src/seismometer/data/pandas_helpers.py
MahmoodEtedadi: FINC.

src/seismometer/data/pandas_helpers.py
MahmoodEtedadi: Updated the code structure to have a function for each aggregation method and moved the logic (if/else blocks) inside aggregation functions.

src/seismometer/data/pandas_helpers.py
MahmoodEtedadi: I think for the analytics table with diabetes data, it went from about 30 seconds to about 3 seconds.
We tested with Sepsis data:
1. Fairness table, with or without `per_context` checkbox it took about 15 seconds. (I think `per_context` took more than a minute before.)
2. `show_cohort_summaries`, takes about 30 seconds.

src/seismometer/data/pandas_helpers.py
MahmoodEtedadi: FINC.

tests/data/test_merge_window.py
MahmoodEtedadi: FINC.

tests/data/test_merge_window.py
MahmoodEtedadi: Added a couple of tests for nearest after and nearest before cases.

changelog/138.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix inconsistent number of positives when scores are combined in ExploreModelEvaluation and show_cohort_summaries. Improve performance of combining scores.
```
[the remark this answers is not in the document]
diehlbw: I cant add a suggestion on this thread now...  
Can you change "fix" to "prevent" -or- delete "inconsistent" (maybe adding something like '... making them more consistent between Explore* and show_*')

src/seismometer/data/pandas_helpers.py
diehlbw: Thanks! I think the separate functions will be easier to keep under (useful) test, not to mention the router makes it clear how to extend

src/seismometer/data/pandas_helpers.py
diehlbw: Resolved; user error **not** an issue.  
and thanks! I knew I was probably overlooking something obvious

tests/data/test_events.py
```diff
@@ -35,6 +35,38 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
+            assert actual["ModelScore"].tolist() == expected_score.tolist()
+        else:
+            actual = undertest.event_score(input_frame, ["Id", "CtxId"], "ModelScore", ref_event, aggregation_method)
+            assert actual["ModelScore"].tolist() == expected_score.tolist()
 
-        actual = undertest.event_score(input_frame, ["Id", "CtxId"], "ModelScore", ref_event, aggregation_method)
-        assert actual["ModelScore"].tolist() == expected_score.tolist()
+    def test_max_aggregation(self, event_data):
+        input_frame, _ = event_data
+        actual = undertest.max_aggregation(input_frame, ["Id", "CtxId"], "ModelScore", "Target")
```
gbowlin, line 47: Can we not reference `event_data` here? 

The reason being that this test should be self contained enough that we don't need a fixture. 

We should only need four scores (two contexts, one with three rows and one with one row) so that we can pick the max from the first three and the only one from the last row) And just the minimal coulmns we need for id, ctx, score, target. 

My distaste for event_data being that its a complicted object and if that object gets updated, out test might break for no good reason.

tests/data/test_merge_window.py
```diff
@@ -907,15 +919,31 @@ def test_merge_first(self):
         actual = actual.sort_values(by=["Id", "CtxId", "PredictTime"]).reset_index(drop=True)
         pdt.assert_frame_equal(actual[expected.columns], expected, check_dtype=False)
 
-    def test_merge_nearest(self):
-        # Test merge_windowed_event with merge strategy equal to first
+    @pytest.mark.parametrize(
```
gbowlin, line 922: This is nice, thanks for sorting this out as its own specific set of params.

src/seismometer/data/pandas_helpers.py
```diff
@@ -437,50 +549,20 @@ def event_score(
...
-        df = merged_frame[merged_frame[event_time].notna()]
-
-    if len(df.index) == 0:
-        return
+    aggregation_methods = {
+        "max": max_aggregation,
+        "min": min_aggregation,
+        "first": first_aggregation,
+        "last": last_aggregation,
+    }
```
gbowlin, line 559: Can we move this to a constant map outside of the `event_score` function?

changelog/138.bugfix.rst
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

tests/data/test_events.py
MahmoodEtedadi: FINC.

tests/data/test_events.py
MahmoodEtedadi: Updated unit tests to be self-contained.

src/seismometer/data/pandas_helpers.py
MahmoodEtedadi: I think the functions are different enough that using same aggregation functions for both needs (probably major) restructuring. Here we are just merging events data into the dataframe and we do not care about duplicates (we are working with `merge_strategy`).
In `event_score` the main objective is to get rid of duplicate entities (contexts) and we are working with `aggregation_method`.

tests/data/test_events.py
```diff
@@ -42,31 +42,72 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
         actual = undertest.max_aggregation(input_frame, ["Id", "CtxId"], "ModelScore", "Target")
-        expected = input_frame.loc[input_frame["ModelScore"] == 21]
+        expected = pd.DataFrame(
+            {"Id": [1, 2], "CtxId": [10, 20], "ModelScore": [2, 1], "Target_Value": [1, 0]}, index=[1, 3]
+        )
         pd.testing.assert_frame_equal(actual, expected)
 
-    def test_min_aggregation(self, event_data):
-        input_frame, _ = event_data
+    def test_min_aggregation(self):
```
gbowlin, line 55: I was thinking something like

```python
    @pytest.mark.parametrize("entities,scores,targets,selected_rows", [
        ([1,1,1,2],[1,2,3,1],[0,1,0,1],[1,3]), # we select the one positive target for both entities
        ([1,1,1,2],[1,2,3,1],[1,1,0,1],[0,3]), # Given two positive targets for the first, we select the smaller score
        ])
    def test_min_aggregation(self, entities, scores, targets, selected_rows):
        input_frame = pd.DataFrame(
            {"Id": entities, "CtxId": [entity*10 for entity in entities], "ModelScore": scores, "Target_Value": targets}
        )
        actual = undertest.min_aggregation(input_frame, ["Id", "CtxId"], "ModelScore", "Target")
        expected =  input_frame.loc[pd.Index(selected_rows)]
        pd.testing.assert_frame_equal(actual, expected)
```

MahmoodEtedadi: FINC.

tests/data/test_events.py
gbowlin: I think the key case here is the event timestamp + windowing code

Something like the example you showed me earlier with  time windows

  encounter  observation   time       prediction
0          1            0               +1hr       threshold+0.2
1          1            1                -1hr       threshold+0.1
2          1            2                -2hrs      threshold-0.1

tests/data/test_events.py
```diff
@@ -42,60 +42,142 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
+        expected = input_frame.loc[pd.Index(selected_rows)]
+        if expected.empty:
+            expected = pd.DataFrame()
         pd.testing.assert_frame_equal(actual, expected)
 
-    def test_last_aggregation(self):
+    @pytest.mark.parametrize(
+        "entities,scores,event_times,selected_rows",
+        [
+            (
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

tests/data/test_events.py
MahmoodEtedadi: Added a few tests to `event_score`.

tests/data/test_events.py
```diff
@@ -35,6 +35,257 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
+        input_frame = pd.DataFrame(
+            {
+                "Id": entities,
+                "CtxId": [entity * 10 for entity in entities],
+                "ModelScore": scores,
+                "Target_Value": targets,
+                "EventTime": pd.to_datetime(event_times),
+            }
+        )
+        ref_event = "Target_Value" if aggregation_method in ["max", "min"] else "EventTime"
```
diehlbw, line 101: Needing this logic worries me a little.  Did our code mess up separation of concerns a bit? ie should this determination be something that event_score (or similar) knows how to do?

Otherwise, is it needed to test this unit of code?

tests/data/test_events.py
```diff
@@ -35,6 +35,257 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
+        actual = undertest.event_score(
+            input_frame, ["Id", "CtxId"], "ModelScore", ref_event, aggregation_method=aggregation_method
+        )
+        expected = input_frame.loc[pd.Index(selected_rows)]
```
diehlbw, line 105: Nit: Can you move this line to before ref_event? 

When possible there is benefit to the Arrange then Act then Assert order for tests.  You can even lean into this expectation by adding a couple lines of whitespace (with or without comments) to suggest the sections.

\# Arrange
input_frame = ..  
expected = ..  
ref_event = .. 

\# Act
actual = ..  

\# Assert 
assert..

Expectation of this pattern is what drew my attention to the ref_event. is that a needed arrange step or would it make sense as part of the act (or would you be having an equivalently good test without it).   

But the reason I think this is worth calling out is that when you set testcase expectations _after_ the act portion, you risk creating tests that have leaked information.  Its more effort to verify this didnt happen than if it was earlier and such things could not occur just because of execution order.

tests/data/test_summaries.py
```diff
@@ -48,16 +52,59 @@ def expected_score_target_summary_cuts(res):
...
+        labels = ["(0,0.5]", "(0.5,1]"]
+
+        # Create a new column for binned scores
+        df_aggregated["Score_Binned"] = pd.cut(
+            df_aggregated["Score"], bins=bins, labels=labels, include_lowest=False, right=True
+        )
+
+        # Group by the desired columns and count
+        entities_event_score = (
+            df_aggregated.groupby(["Has_ECG", "Target_Value", "Score_Binned"])
```
diehlbw, line 100: I'm seeing a warning from this line in my pyenv, needs an explicit set of `observed`
```suggestion
            df_aggregated.groupby(["Has_ECG", "Target_Value", "Score_Binned"], observed=False)
```

tests/data/test_events.py
```diff
@@ -35,6 +35,257 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
+            {
+                "Id": entities,
+                "CtxId": [entity * 10 for entity in entities],
+                "ModelScore": scores,
+                "EventTime": pd.to_datetime(event_times),
+            }
+        )
+        actual = undertest.last_aggregation(input_frame, ["Id", "CtxId"], "ModelScore", "EventTime")
+        expected = input_frame.loc[pd.Index(selected_rows)]
+        if expected.empty:
```
diehlbw, line 242: Again using the test to question the code.  
Would it make more sense if first/last returned a zero-row frame instead of the new null one?

I don't think any of the other aggregations would be expected to alter columns, so why should this edge case?   It seems like 0-row frames don't cause issue with sort and indexing either.

tests/data/test_events.py
MahmoodEtedadi: FINC.

tests/data/test_events.py
MahmoodEtedadi: For first/last we used to return empty DataFrame instead of the same dataframe with no rows (columns are different). Updated to return dataframe with no rows and removed the if statement from the unit test.

tests/data/test_summaries.py
```diff
@@ -48,16 +52,59 @@ def expected_score_target_summary_cuts(res):
...
+    @patch.object(seismogram, "Seismogram", return_value=Mock())
+    @pytest.mark.parametrize("aggregation_method", ["min", "max", "first", "last"])
+    def test_event_score_match_score_target_summaries(
+        self, mock_seismo, aggregation_method, prediction_data, expected_score_target_summary_cuts
+    ):
+        fake_seismo = mock_seismo()
+        fake_seismo.output = "Score"
+        fake_seismo.target = "Target"
+        fake_seismo.event_aggregation_method = lambda x: aggregation_method
+        ref_event = "Target"
```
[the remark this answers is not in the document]
MahmoodEtedadi: Grouped lines associated with each approach together.

tests/data/test_events.py
```diff
@@ -26,23 +26,27 @@ def event_data(res):
 # fmt: off
 
 class Test_Event_Score:
-    @pytest.mark.parametrize("ref_event", ["Target", "PredictTime", "Reference_5_15_Time"])
+    @pytest.mark.parametrize(
+            "ref_time,ref_event", [("Target", "Target"), ("PredictTime", "Target"), ("Reference_5_15_Time", "Target")]
```
diehlbw, line 30: Instead of ref_col logic, can it instead be added here as expected_col

tests/data/test_events.py
```diff
@@ -35,6 +35,257 @@ def test_bad_event(self, id_, ctx, aggregation_method, ref_event, event_data):
...
+        if ref_event in ["PredictTime", "Reference_5_15_Time"] and aggregation_method in ["min", "max"]:
+            actual = undertest.event_score(input_frame, ["Id", "CtxId"], "ModelScore", "Target", aggregation_method)
+            assert actual["ModelScore"].tolist() == expected_score.tolist()
+        else:
+            actual = undertest.event_score(input_frame, ["Id", "CtxId"], "ModelScore", ref_event, aggregation_method)
+            assert actual["ModelScore"].tolist() == expected_score.tolist()
 
-        actual = undertest.event_score(input_frame, ["Id", "CtxId"], "ModelScore", ref_event, aggregation_method)
-        assert actual["ModelScore"].tolist() == expected_score.tolist()
+    @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

tests/data/test_events.py
MahmoodEtedadi: Added `ref_time` as a param to `event_score` and updated unit tests accordingly.

tests/data/test_events.py
diehlbw: Chatted and this makes it less understandable

diehlbw (approved):

timotheois:
Testing notes say "per_target=True, per_score=True" but the parameters seem to be "by_target" and "by_score" - may be a typo

MahmoodEtedadi:
Yes, I have updated instructions. It should be "by_target" and "by_score".

diehlbw (approved):

gbowlin (approved):

PR #149, opened 2025-04-28
Add info/debug logging to track how predictions/events data are changed

MahmoodEtedadi:
# Overview
Adds general logging across Seismometer to improve visibility
into data handling and processing.

## Description of changes
Added additional INFO and DEBUG logs throughout data loading, filtering,
merging, and evaluation to capture row counts and event handling. Updated
the notebook and User Guide to document `log_level` usage (`WARNING/30`
by default, use `INFO/20` or `DEBUG/10` for more detail).


firste:
I would like to mention the logging support in our docs somewhere more visible than the autogenerated API details. At a minimum can you please update the example notebook with explicitly setting the log level to the default? I expect most users will start from the example notebook so this is the right place to catch folks. The FAQ would also be a great place to add more details about troubleshooting.

firste (changes requested):
Please update the documentation.

MahmoodEtedadi:
> Please update the documentation.

added a logging section to the documentation and `log_level` param in the example notebook.

PR #153, opened 2025-05-02
Improve accuracy of AUC computation in analytics table

### LINKED ISSUE #141: ROC AUC value differs between ExploreModelEvaluation() and ExploreAnalyticsTable()
## Problem Summary
ROC AUC value differs between ExploreModelEvaluation() and ExploreAnalyticsTable()

## Impact

Though, the ROC AUC value is similar, and may not be operationally meaningful, the difference is likely to decrease trust in the values output by this package (for people such as myself).
## Possible Solution
It appears the annotation value in the ROC AUC curve in ExploreModelEvaluation() is most consistent with the scikit-learn roc_auc function, using decimal thresholds 1-100 whereas the ExploreAnalyticsTable() is calculated from a dataframe with integer thresholds 1-100. I believe this accounts for the difference.

## Steps to Reproduce

~~~python
import numpy as np
import matplotlib.pyplot as plt
from sklearn.metrics import roc_curve, auc
sg = sm.Seismogram()

# derive comparison dataframes from the Seismogram object and calculate_bin_stats method
df = sg.dataframe
stats = sm.data.performance.calculate_bin_stats(y_true=df['Sepsis within 8 hours_Value'], y_pred=df['EPICSEPSISSCORE'])

# calculate the ROC AUC using sklearn
y_true=df['Sepsis within 8 hours_Value']
y_pred=df['EPICSEPSISSCORE']

fpr, tpr, thresholds = roc_curve(y_true, y_pred)

roc_auc_sklearn = auc(fpr, tpr)

#calculate the ROC AUC using the stats dataframe
stats_auc = auc(1 - stats["Specificity"], stats["Sensitivity"])

#print the results revelaing the difference between the two methods
print(roc_auc_sklearn.round(2))
print(stats_auc.round(2))
print((roc_auc_sklearn - stats_auc).round(3))
~~~
## Suggested fix
Use the scikit-learn methodology to produce roc auc for ExploreAnalyticsTable()

MahmoodEtedadi:
# Overview
We generally use an approximation with 101 threshold points to estimate AUC. In plots like `ExploreModelEvaluation`, where values are displayed with only two decimal places, this approximation is sufficient. However, in `ExploreAnalyticsTable`, which displays values to three decimal places, the difference between using 101 points and a full-resolution calculation becomes more noticeable.
Closes #141 

## Description of changes
To improve the accuracy of AUC in `ExploreAnalyticsTable` while maintaining performance, we now use 1001 threshold points instead of 101. This offers a better balance between precision and efficiency, especially given the increased decimal display precision.

## Author Checklist

### DIFF
--- a/src/seismometer/data/performance.py
+++ b/src/seismometer/data/performance.py
@@ -164,7 +164,7 @@ class BinaryClassifierMetricGenerator(MetricGenerator):
         stats = stats.loc[score_threshold_integer]
         return stats.to_dict()
 
-    def calculate_binary_stats(self, dataframe, target_col, score_col, metrics):
+    def calculate_binary_stats(self, dataframe, target_col, score_col, metrics, threshold_precision=0):
         """
         Calculates binary stats for all thresholds.
 
@@ -178,10 +178,20 @@ class BinaryClassifierMetricGenerator(MetricGenerator):
             The column in the dataframe that contains the predicted scores.
         metrics: list[str], optional
             List of metrics to filter down to.
+        threshold_precision : int, optional
+            Number of decimal places to use when generating thresholds as percentages.
+            - E.g., `threshold_precision=0` yields thresholds like 0, 1, ..., 100 (coarse).
+            - `threshold_precision=2` yields 0.00, 0.01, ..., 100.00 (fine-grained).
+            - Higher values improve AUC approximation but increase computation cost.
+            By default 0.
         """
         y_true = dataframe[target_col]
         y_pred = dataframe[score_col]
-        stats = calculate_bin_stats(y_true, y_pred, rho=self.rho).round(5).set_index(THRESHOLD)
+        stats = (
+            calculate_bin_stats(y_true, y_pred, rho=self.rho, threshold_precision=threshold_precision)
+            .round(5)
+            .set_index(THRESHOLD)
+        )
         for name, percent in zip(COUNTS, PERCENTS):
             stats[percent] = stats[name] * 100.0 / len(dataframe)
 
@@ -242,6 +252,7 @@ def calculate_bin_stats(
     keep_score_values: bool = False,
     not_point_thresholds: bool = False,
     rho: float = None,
+    threshold_precision: int = 0,
 ) -> pd.DataFrame:
     """
     Calculate summary statistics from y_true and y_pred (y_proba[:,1] for binary classification) arrays.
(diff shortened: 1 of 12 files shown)


src/seismometer/data/performance.py
```diff
@@ -178,10 +178,20 @@ def calculate_binary_stats(self, dataframe, target_col, score_col, metrics):
...
+            - `threshold_precision=2` yields 0.00, 0.01, ..., 100.00 (fine-grained).
+            - Higher values improve AUC approximation but increase computation cost.
+            By default 0.
         """
         y_true = dataframe[target_col]
         y_pred = dataframe[score_col]
-        stats = calculate_bin_stats(y_true, y_pred, rho=self.rho).round(5).set_index(THRESHOLD)
+        stats = (
+            calculate_bin_stats(y_true, y_pred, rho=self.rho, threshold_precision=threshold_precision)
+            .round(5)
```
gbowlin, line 192: Since we now have a `threshold_precision`, should we use that to set the rounding (or more the rounding into `calculate_bin_stats`?

tests/data/test_binary_performance.py
```diff
@@ -247,3 +247,52 @@ def test_per_context_does_not_leak_state_across_iterations(self, fake_seismo):
...
+
+        for col, atol in [("0.5_Threshold", 0.1), ("AUROC", 0.001)]:
+            assert col in result_low.columns and col in result_high.columns
+
+            val_low = result_low[col].iloc[0]
+            val_high = result_high[col].iloc[0]
+
+            # Values differ (better precision affects threshold resolution)
+            assert val_low != val_high
+
```
gbowlin, line 296:
```suggestion
            # Values differ (better precision affects threshold resolution)
            assert val_low != val_high, "These can collide with probability 1:100"
```
Really since we have a for loop, the probability that this fails is `1 - (.99)^2` which is close to 1:50
Can we not use RNG?

If we want to use RNG and we want to automatically rerun on failure, there are some plugins for this:
https://docs.pytest.org/en/stable/explanation/flaky.html has some suggestions if we wanted to go that route (https://github.com/pytest-dev/pytest-rerunfailures?tab=readme-ov-file#re-run-individual-failures)

I don't think that extra work is needed, and an assertion comment should suffice.

gbowlin: We could also drop the assert that they are not equal.

src/seismometer/data/performance.py
gbowlin: Never mind, I didn't see the rounding is on the stats, not the thresholds.

gbowlin (changes requested):
Can we fix the unit test being probabalistic?

tests/data/test_binary_performance.py
gbowlin: I didn't catch the seed of `42` up top in `rng = np.random.default_rng(42)`. 

I still think an assertion comment would be useful. Something like 

```suggestion
            # Values differ (better precision affects threshold resolution)
            assert val_low != val_high, "Differing precision expected, test seed was chosen to verify these two values are not equal"

```

gbowlin (approved):

PR #154, opened 2025-05-02
Combine consistent metric values

MahmoodEtedadi:
# Overview

Closes #xxx 

## Author Checklist

### DIFF
--- a/.devcontainer/image/Dockerfile
+++ b/.devcontainer/image/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11-slim as seismometer
+FROM python:3.12-slim as seismometer
 
 RUN useradd --create-home --uid 1000 --shell /bin/bash seismo
 ENV PATH=${PATH}:/home/seismo/.local/bin
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.12'
       - name: Install towncrier
         run: |
           pip install --upgrade pip
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.12"
       - name: Install pre-commit
         run: |
           pip install --upgrade pip
@@ -27,7 +27,7 @@ jobs:
         run: |
           pre-commit run --all-files
   test-coverage:
-    name: Coverage py3.10
+    name: Coverage py3.12
     runs-on: ubuntu-latest
 
     steps:
@@ -36,7 +36,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.12"
       - name: Install dependencies
         run: |
           pip install --upgrade pip
@@ -69,7 +69,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.12"
       - name: Install dependencies
         run: |
           pip install --upgrade pip
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10' 
+          python-version: '3.12' 
       - name: Install Graphviz / Pandoc
         run: |
           sudo apt-get install graphviz libgraphviz-dev pandoc
--- a/changelog/147.bugfix.rst
+++ b/changelog/147.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug in `generate_analytics_data` where `per_context=True` caused data contamination across iterations due to in-place modification of the input DataFrame.
--- a/changelog/150.bugfix.rst
+++ b/changelog/150.bugfix.rst
@@ -1,0 +1,1 @@
+Limit ``pyarrow`` version to ``>14.0.0,<20.0.0`` as ``use_legacy_dataset`` keyword is removed in later versions.
--- a/changelog/152.misc.rst
+++ b/changelog/152.misc.rst
@@ -1,0 +1,1 @@
+Updated GitHub Actions workflow to target Python 3.12.
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ install_requires =
     seaborn>=0.13.2,<0.14
     pandas>=2.1,<3
     scikit-learn>=1.2.0,<2
-    pyarrow>=9.0.0
+    pyarrow>14.0.0,<20.0.0
     pydantic>=2.6.3,<3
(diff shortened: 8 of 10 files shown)


src/seismometer/controls/categorical.py
```diff
@@ -86,15 +93,66 @@ def _extract_metric_values(self):
...
+                    in_degree[b] += 1
+                in_degree.setdefault(a, 0)
+
+        result = []
+
+        # Step 2: Repeatedly find the only valid next node
+        while len(result) < len(all_nodes):
+            candidates = [n for n in all_nodes if in_degree[n] == 0 and n not in result]
+
+            if len(candidates) != 1:
```
gbowlin, line 146: candidates will be undefined if all_nodes is empty (dumb case where there is only one category value to count for each list being merged.)

gbowlin (changes requested):

src/seismometer/controls/categorical.py
MahmoodEtedadi: I think `all_nodes` is empty iff all value lists are empty:
```python
for lst in lists:
            all_nodes.update(lst)
```
and it is defined if `all_nodes` is not empty. If all lists are empty, then the data generated will be an empty dataframe -> "plot censored" message.

Also, if there are lists of length 1 and: 
i. values are the same: then `all_nodes` has the 1 value, canididates will be the 1 valueas well and we quit after one iteration
ii. values are different: then candidates has more than 1 values and we fail because of ambiguity.

src/seismometer/controls/categorical.py
```diff
@@ -86,15 +93,66 @@ def _extract_metric_values(self):
...
+                raise ValueError("Inconsistent or ambiguous ordering: cannot determine a unique next value.")
+
+            node = candidates[0]
+            result.append(node)
 
-        if len(first) > MAX_CATEGORY_SIZE:
-            raise ValueError(f"Total number of values ({len(first)}) exceeds MAX_CATEGORY_SIZE ({MAX_CATEGORY_SIZE}).")
+            for neighbor in graph[node]:
```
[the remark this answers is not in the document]
MahmoodEtedadi: Deleting the node here as it will not be used later.

src/seismometer/controls/categorical.py
```diff
@@ -86,15 +93,66 @@ def _extract_metric_values(self):
...
+            If there are conflicting or ambiguous orderings.
+        """
+        graph = defaultdict(set)
+        in_degree = defaultdict(int)
+        all_nodes = set()
+
+        # Step 1: Build graph edges from each list
+        for lst in lists:
+            all_nodes.update(lst)
+            for a, b in zip(lst, lst[1:]):
```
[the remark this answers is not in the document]
MahmoodEtedadi: We could have [1], [3], [5], [1,3,5]. I think we need to go over all lists before we can identify isolated nodes. And after going over all lists, the first iteration in the while loop checks that we do not have more than one isolated node.

src/seismometer/controls/categorical.py
```diff
@@ -86,15 +93,70 @@ def _extract_metric_values(self):
...
+
+        result = []
+
+        # Step 2: Find total ordering by returning a sequence of directed graph sources (the least element in
+        #         each ordered graph)
+        while len(result) < len(all_nodes):
+            candidates = [n for n in all_nodes if in_degree[n] == 0 and n not in result]
+
+            if len(candidates) != 1:
+                raise ValueError("Inconsistent or ambiguous ordering: cannot determine a unique next value.")
```
gbowlin, line 148: raise ValueError("Inconsistent or ambiguous ordering: cannot determine a unique next value.")
ValueError: Inconsistent or ambiguous ordering: cannot determine a unique next value.

Can we tell people which part of the config is ambiguous in the error, so that I can fix the configuration?

I'm seeing this in the fairness widget and it doesnt help with debugging where my configuration is wrong;

image

gbowlin (changes requested):
There is also a bug when we add a new ordinal feedback column without data that we should write up as an issue and fix (just filter out columns that don't exist - we already have a warning for it during loading)

src/seismometer/controls/categorical.py
```diff
@@ -84,17 +91,229 @@ def _extract_metric_values(self):
...
+
+        # Step 1: Build directed graph from the order of each list
+        for metric in metric_to_values:
+            lst = metric_to_values[metric]
+            all_nodes.update(lst)
+            for a, b in zip(lst, lst[1:]):
+                if b not in graph[a]:
+                    graph[a].add(b)
+                    in_degree[b] += 1
+                in_degree.setdefault(a, 0)
```
diehlbw, line 143: What does setdefault for a defaultdict?
I thought one of the common benefits of defaultdict is to obviate this need.

If you need to "touch" in order to update the keys, you might just do a `_=in_degree[a]` or... perhaps you can just use a normal dict.

src/seismometer/controls/categorical.py
```diff
@@ -84,17 +91,229 @@ def _extract_metric_values(self):
...
+                idx = path.index(node)
+                return path[idx:] + [node]
+
+            if node in visited:
+                return None
+
+            visited.add(node)
+            path.append(node)
+
+            for neighbor in graph.get(node, []):
```
diehlbw, line 303: Is dfs knowing about `graph` due to being nested intentional?   

I can't fully think through implications on the scoping of this variable within a nested inner function.  Should we be worried or is it not a problem?

diehlbw (changes requested):

src/seismometer/controls/categorical.py
```diff
@@ -84,17 +91,229 @@ def _extract_metric_values(self):
...
+        ------
+        ValueError
+            If there are conflicting or ambiguous orderings.
+        """
+        graph = defaultdict(set)
+        in_degree = defaultdict(int)
+        all_nodes = set()
+
+        # Step 1: Build directed graph from the order of each list
+        for metric in metric_to_values:
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

src/seismometer/controls/categorical.py
MahmoodEtedadi: Removed `setdefault` and changed `del ...` to `in_degree.pop(node, None)` just to be safe. (I think `candidates = [n for n in all_nodes if in_degree[n] == 0 and n not in result]` initializes `in_degree` for all nodes but just to be safe change `del` to `.pop`.)

src/seismometer/controls/categorical.py
MahmoodEtedadi: Yes, since `dfs` is a generic function to find a cycle (if one exists) in a graph and `_find_any_cycle` is a private function, we probably can leave the structure as is. 
As far as variable scoping, I think `graph` is available as a read-only variable inside the function and we do not change it in `dfs`. (`graph.get(node, [])` does not modify `graph` even if it is `defaultdict`.)

src/seismometer/controls/categorical.py
```diff
@@ -84,17 +91,229 @@ def _extract_metric_values(self):
...
+
+        Returns
+        -------
+        list[str]
+            A list of nodes forming a cycle, with the start node repeated at the end to close the loop.
+            Returns an empty list if no cycle is found (should not happen if a cycle is guaranteed).
+        """
+        visited = set()
+
+        def dfs(node, path):
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

diehlbw (approved):

Issue #130, opened 2025-02-06
Add more unit tests for pandas_helpers


MahmoodEtedadi:
## Is your feature request related to a problem? Please describe
Given that subtle changes in pandas behavior could impact how event windows are handled, it would be prudent to add more comprehensive unit tests for the relevant functions.

## Describe the solution you'd like
Adding more unit tests for pandas_helpers (in particular, merge_windowed_event) would be helpful.

gbowlin:
While #155 added a fair number of tests to pandas helpers, we still want to improve the quality of unit tests. 

In particular - we should try to come up with minimal data frame examples for each test that are easy to  validate by hand. This way we can be sure that tests only cover the data they need to run, and that the example makes sense given the name of the test.

PR #155, opened 2025-05-06
Increase test coverage

MahmoodEtedadi:
# Overview
Increase unit test coverage to >94.

## Author Checklist


tests/api/test_plots.py
```diff
@@ -0,0 +1,449 @@
...
+def get_test_loader(config):
+    mock_loader = Mock(autospec=SeismogramLoader)
+    mock_loader.config = config
+
+    return mock_loader
+
+
+def get_test_data():
+    return pd.DataFrame(
+        {
```
gbowlin, line 98: nit: Can you explain why we want to use this as test data across all our tests? What features does it have that help each of the tests below?

Can we/should we have different data for different tests?

tests/api/test_plots.py
```diff
@@ -0,0 +1,449 @@
...
+        result = cohort_list_details(fake_seismo.available_cohort_groups)
+        assert isinstance(result, HTML)
+        assert "Summary" in result.data
+
+
+class TestExploreBinaryModelMetrics:
+    def test_exploration_metric_widget_initialization(self, fake_seismo):
+        widget = ExploreBinaryModelMetrics(rho=0.5)
+        assert widget.metric_generator.rho == 0.5
+        assert "Model Metric Evaluation" in widget.children[0].value
```
gbowlin, line 182: Should this test be in a different file?

tests/api/test_plots.py
```diff
@@ -0,0 +1,449 @@
...
+        df = fake_seismo.dataframe
+        df["event1_Value"] = 1  # no variation
+        html = _model_evaluation(df, ["entity"], "event1", "event1_Value", "score1", [0.2], 0)
+        assert "requires exactly two classes" in html.data
+
+
+class TestExploreModelScoreComparison:
+    def test_generate_plot(self, fake_seismo):
+        self.widget = ExploreModelScoreComparison()
+        result = self.widget._try_generate_plot()
```
gbowlin, line 217: If we are looking at code coverage, I'd rather not directly test `_try_generate_plot()` for a given subclass as it will call a LOT of non-mocked downstream code, which will push our coverage number up in a non-meaningful way.

If we have code that it doesn't make sense to test, maybe exempt it at the right level?

gbowlin (changes requested):
Just a question on the unit tests around calling `_try_generate_plot`

tests/api/test_plots.py
MahmoodEtedadi: Yes, moved to `test_explore.py`.

tests/api/test_plots.py
MahmoodEtedadi: Here we test various explore classes in `api/explore.py` to make sure they produce an appropriate output without errors. We need to have data similar to `sg.dataframe` for all use cases. 
We do not have tests explicitly for various plots in `plots.py`. Changed the name of the test file to `test_api_explore.py`

tests/api/test_plots.py
MahmoodEtedadi: Updated tests to not directly call `_try_generate_plot()`. Added tests for various plot functions.

tests/api/test_api_explore.py
```diff
@@ -176,19 +185,105 @@ def test_cohort_list_details_single_target_index_rename(self, fake_seismo):
...
+        metric_gen = BinaryClassifierMetricGenerator(rho=0.5)
+        html = plot_binary_classifier_metrics(
+            metric_generator=metric_gen,
+            metrics=["Accuracy", "PPV"],
+            cohort_dict={},
+            target="event1",
+            score_column="score1",
+        )
+        assert isinstance(html, SVG)
+        assert "Accuracy" in html.data and "PPV" in html.data
```
gbowlin, line 231: I worry about this style of test (generating a plot from data) being too invasive, meaning we get free code coverage for "not crashing", not for giving the right answer. Can we mock how the plot generation? If not we can just remove these.

gbowlin: Can you do the same in other places as well

tests/data/test_pandas_helpers.py
```diff
@@ -47,6 +47,20 @@ def onekey(df):
     return CaseData(*(filter_fn(element) for element in vars(casedata).values()))
 
 
+@pytest.fixture
+def base_counts_data():
```
gbowlin, line 51: Can we give a better description of this, or why its being used in specific tests.

tests/data/test_pandas_helpers.py
```diff
@@ -258,3 +293,483 @@ def test_three_suffixes_align(self, base, suffix):
...
+                "MyEvent",
+                "Label",
+                window_hrs=5,
+            )
+
+        # Ensure truncation happened
+        assert len([c for c in result.columns if "_Count" in c]) == undertest.MAXIMUM_COUNT_CATS
+        assert "Maximum number of unique events" in caplog.text
+
+    def test_counts_are_correct(self, base_counts_data):
```
gbowlin, line 520: Maybe use a specific dataframe for just this test to make the use case more clear?

tests/data/test_pandas_helpers.py
```diff
@@ -258,3 +293,483 @@ def test_three_suffixes_align(self, base, suffix):
...
+            {
+                "Id": [1, 1],
+                "Time": [
+                    pd.Timestamp("2024-01-01 03:00:00"),
+                    pd.Timestamp("2024-01-01 05:00:00"),
+                ],
+                "Value": [1, 1],
+                "Type": ["MyEvent", "MyEvent"],
+            }
+        )
```
gbowlin, line 609: nit: explain what is going on with these dataframes (user expectation, not code expectation) (what should be ignored, what should be expected)

tests/data/test_pandas_helpers.py
```diff
@@ -258,3 +293,483 @@ def test_three_suffixes_align(self, base, suffix):
...
+        assert result["Label~A_Count"].iloc[0] == 1
+        assert result["Label~B_Count"].iloc[0] == 1
+
+        # ID 2 should get zero counts
+        assert result["Label~A_Count"].iloc[1] == 0
+        assert result["Label~B_Count"].iloc[1] == 0
+
+
+class TestMergeWindowedEvent:
+    def test_basic_forward_strategy(self):
```
gbowlin, line 588: Is this test a duplicate of the parametrized one (that comes next)?

tests/data/test_pandas_helpers.py
```diff
@@ -258,3 +293,483 @@ def test_three_suffixes_align(self, base, suffix):
...
+            event_label="MyEvent",
+            pks=["Id"],
+            min_leadtime_hrs=0,
+            window_hrs=3,
+            merge_strategy="count",
+            event_base_val_col="Value",
+            event_base_time_col="Time",
+        )
+
+        assert "MyEvent~A_Count" in result.columns
```
gbowlin, line 711: Do we need to test counts?

tests/data/test_pandas_helpers.py
```diff
@@ -258,3 +293,483 @@ def test_three_suffixes_align(self, base, suffix):
...
+                min_leadtime_hrs=0,
+                window_hrs=5,
+                merge_strategy="forward",
+                event_base_val_col="Value",
+                event_base_time_col="Time",
+            )
+
+
+def test_post_process_event_skips_cast_when_dtype_none():
+    df = pd.DataFrame({"Label": [None], "Time": [pd.Timestamp.now()]})
```
gbowlin, line 758: Shouldn't testing a skip make sure some method is not called? WIth assert.not_called or something?

gbowlin (changes requested):
Along the same lines as before, can we make sure that any unit tests we add are scoped correctly to not produce excess code coverage without any asserts to make sure those results are good? 

We want to avoid tests that run a lot of code without testing that code.

tests/api/test_api_explore.py
gbowlin: Mocking `metric_gen`  is all you need here.

tests/api/test_api_explore.py
```diff
@@ -0,0 +1,666 @@
...
+class TestExploreModelScoreComparison:
+    def test_widget_calls_plot_function(self, fake_seismo):
+        widget = ExploreModelScoreComparison()
+
+        widget.option_widget.cohort_list.value = {"Cohort": ["C1", "C2"]}
+        widget.option_widget.model_options.target_list.value = "event1"
+        widget.option_widget.model_options.score_list.value = ("prediction", "score1")
+        widget.option_widget.model_options.per_context_checkbox.value = False
+
+        result = widget.plot_function(
```
gbowlin, line 318: Can we just mock the plot function everywhere you care calling it? I think that's the real result you want, isnt it? Then you can just assert that its called with the right options.

tests/api/test_api_explore.py
MahmoodEtedadi: Mocked `BinaryClassifierMetricGenerator.calculate_binary_stats` as this is the function that in the tests path goes outside scope (`plots.py`).

tests/api/test_api_explore.py
MahmoodEtedadi: Updated tests to mock (almost) every call to outside `explore.py` or `plots.py`.

tests/data/test_pandas_helpers.py
MahmoodEtedadi: Updated.

tests/api/test_api_explore.py
```diff
@@ -299,235 +331,279 @@ def dummy_plot(*args, **kwargs):
...
-            plots.plot, "cohorts_vertical", lambda *_a, **_k: (_ for _ in ()).throw(ValueError("fail"))
-        )
-        html = plots._plot_cohort_hist(fake_seismo.dataframe, "event1_Value", "score1", "Cohort", ["C1", "C2"], 0)
-        assert "Error" in html.data
+    @patch("seismometer.html.template.render_title_message", return_value=HTML("error"))
+    @patch("seismometer.api.plots.plot.cohorts_vertical", side_effect=ValueError("fail"))
+    def test_plot_cohort_hist_plot_fails(self, mock_plot, mock_render, fake_seismo):
+        df = fake_seismo.dataframe.copy()
+        result = _plot_cohort_hist(df, "event1_Value", "score1", "Cohort", ["C1", "C2"])
+        assert "error" in result.data.lower()
```
gbowlin, line 555: Is this test useful, should it be removed?

tests/api/test_api_explore.py
```diff
@@ -537,130 +613,160 @@ def test_leadtime_enc_missing_target_zero_column(self, fake_seismo, caplog):
...
+            score="score1",
+            threshold=0.2,
+            ref_time="time",
+            cohort_col="Cohort",
+            subgroups=["C1"],
+            max_hours=48,
+            x_label="Lead Time (hours)",
             censor_threshold=5,
         )
         assert "censored" in result.data.lower()
```
gbowlin, line 668: Is this test needed? the next test is hitting the censor limit as well

tests/api/test_api_explore.py
```diff
@@ -537,130 +613,160 @@ def test_leadtime_enc_missing_target_zero_column(self, fake_seismo, caplog):
...
             cohort_col="Cohort",
             subgroups=["C1", "C2"],
             outcome="outcome1",
             intervention="intervention1",
             reference_time_col="time",
+            censor_threshold=0,
         )
-
+        assert isinstance(result, HTML)
+        assert "wrapped outcome/intervention" in result.data
```
gbowlin, line 710: Maybe add check that mock_plot was called

gbowlin (changes requested):
Smaller fixes and checking to see if there are duplicates

tests/api/test_api_explore.py
MahmoodEtedadi: Removed it.

tests/api/test_api_explore.py
MahmoodEtedadi: Here we are trying to hit:
```python
bins = np.histogram_bin_edges(cData["pred"], bins=20)
    try:
        svg = **plot.cohorts_vertical**(cData, plot.histogram_stacked, func_kws={"show_legend": False, "bins": bins})
        title = f"Predicted Probabilities by {cohort_col}"
        return template.render_title_with_image(title, svg)
    except Exception as error:
        return template.render_title_message("Error", f"Error: {**error**}")

File: tests/api/test_api_explore.py [hunk 15]
```diff
@@ -299,235 +331,279 @@ def dummy_plot(*args, **kwargs):
...
-        result = widget.plot_function(
-            cohort_col="Cohort",
-            subgroups=["C1", "C2"],
-            target_column="event1",
-            score_column="score1",
+    @patch("seismometer.html.template.render_censored_plot_message", return_value=HTML("censored"))
+    @patch("seismometer.api.plots.assert_valid_performance_metrics_df", side_effect=ValueError("invalid"))
+    @patch("seismometer.api.plots.get_cohort_performance_data")
+    @patch("seismometer.api.plots.pdh.get_model_scores")
+    def test_plot_cohort_evaluation_invalid_data(
```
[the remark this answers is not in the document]
MahmoodEtedadi: Updated.

tests/data/test_pandas_helpers.py
MahmoodEtedadi: Will update in subsequent PRs.

tests/data/test_pandas_helpers.py
MahmoodEtedadi: Will update in subsequent PRs.

tests/data/test_pandas_helpers.py
```diff
@@ -77,6 +78,52 @@ def test_merge_earliest(self, id_, enc, merge_data):
         # check_like = ignore column order
         pd.testing.assert_frame_equal(actual.reset_index(drop=True), expect, check_like=True, check_dtype=False)
 
+    @pytest.mark.parametrize("strategy", ["forward", "nearest", "first", "last"])
+    def test_merge_with_strategy_variants(self, strategy):
```
gbowlin, line 82: Can we remove this test, it looks like all we are testing is that if we start with two rows, we finish with two rows. This might be more useful if we had events be more than 2 events, in which case the selection down to only two rows is actually limiting.  (In which case, the test name should be `test_merge_strategies_do_not_generate_additional_rows`

Suggestion, rename, and add two events,  an event before 1, and an event at 3

Or just delete this test.

tests/data/test_pandas_helpers.py
gbowlin: Adding these to https://github.com/epic-open-source/seismometer/issues/130

gbowlin (approved):

PR #159, opened 2025-05-28
Support reading CSV and TSV files as data input

eptgen:
# Overview

Currently, seismometer has no support for importing CSV and TSV files as predictions or events. This pull request adds and tests this functionality.

## Description of changes

The user can now import CSV and TSV files as predictions and events, and mix file types as well--for example, the user can import predictions as a CSV file and events as a Parquet file. Seismometer decides what file type the datasets are by their file extension designated in the config file. By default, seismometer assumes that the datasets are Parquet files. There is a difference in how Parquet files and CSV files interpret data types, so during the import process for CSVs and TSVs, we ensure that we cast columns shared between predictions and events, and whose type is defined as "object", into strings.

Recommended testing:
- [ ] Open the notebook contained in `example-notebooks/binary-classifier/classifer_bin.ipynb`.
- [ ] If using your own dataset, populate the directory with the relevant configuration and dataset files and move on. To use the testing dataset:
     - [ ] Import the diabetes dataset once using `download_example_dataset` in the notebook; make sure you don't run the import process again, or it will break the following steps.
     - [ ] Convert the `data/predictions.parquet` and `data/events.parquet` files into CSV and TSV files named `data/predictions.csv`, etc. If doing this with pandas, make sure to use the `index=False` argument to ensure you don't add extra columns. **Note: ** if you don't see the `data` directory, reload the file explorer.
- [ ] Change `prediction_path` and `event_path` in `config.yml` to some combination of CSV, TSV, and Parquet files (as long as they're not both Parquet files).
- [ ] For ease of retesting, add the `reset=True` argument to `run_startup` in the notebook.
- [ ] Run the code block with `run_startup`.
- [ ] Run the rest of the codeblocks in the notebook to make sure everything runs correctly.
     - [ ] If there is confusion whether a codeblock is running as expected, compare the outputs to the output when making the predictions and events datasets both Parquet files in the config.

## Author Checklist


tests/data/loaders/test_load_events.py
```diff
@@ -109,6 +150,10 @@ def test_load_remaps_columns(self, setup_fn, load_fn):
 
         actual = load_fn(config).sort_index(axis=1)
 
+        # the post_transform_fn will convert the Time column to a datetime64[ns] type,
+        # so we need to mock that functionality
+        actual["Time"] = actual["Time"].astype("<M8[ns]")
+
```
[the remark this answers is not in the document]
eptgen: I don't think we should do it in the loader. One thing I noticed while doing development is that the Parquet files we're using has the correct types in the schema already, which isn't the case in the case of CSV files, which is why we haven't needed this cast until now. We already take care of all the typing in the post-load functions (post_transform_fn in event.py and dictionary_types in prediction.py)--in fact, that seems like one of the main purposes of these functions--so it feels redundant to change types in the loader as well.

src/seismometer/data/loader/__init__.py
```diff
@@ -23,9 +40,22 @@ def loader_factory(config: ConfigProvider, post_load_fn: ConfigFrameHook = None)
     SeismogramLoader
         Instance of the SeismogramLoader class, with load_data() method ready to be called
     """
+
+    # Default filetype: parquet
     prediction_loader: ConfigOnlyHook = prediction.parquet_loader
     event_loader: ConfigOnlyHook = event.parquet_loader
```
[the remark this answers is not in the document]
eptgen: I changed the logic to happen in the loaders instead of the constructor.

src/seismometer/data/loader/prediction.py
```diff
@@ -94,7 +159,8 @@ def dictionary_types(config: ConfigProvider, dataframe: pd.DataFrame) -> pd.Data
         If any columns cannot be converted to the expected types.
     """
 
-    defined_types = _gather_defined_types(config)
+    defined_types = gather_prediction_types(config)
+    logger.debug("defined_types: " + str(defined_types))
```
diehlbw, line 163: Nit: Is this log statement intended to be committed?  If so, can it be more descriptive - "Defined types: " vs direct varname "defined_types"

src/seismometer/data/loader/utils.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
```
diehlbw, line 1: I'm obligated to ask is "utils" really a module we want to create?  
The hesitation is that these become a dumping ground for functions, that said I'm not thinking of any significantly better options

src/seismometer/data/loader/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+from seismometer.configuration import ConfigProvider
+
+
+def gather_prediction_types(config: ConfigProvider) -> dict[str, str]:
```
diehlbw, line 6: We own ConfigProvider, should this just be added as a nice accessor there maybe even as a property?

`config.prediction_types`

src/seismometer/data/loader/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    -------
+    dict[str, str]
+        the type dictionary with the structure dict[column] = type
+    """
+    return {defn.name: defn.dtype for defn in config.prediction_defs.predictions if defn.dtype is not None}
+
+
+def get_loader_from_path(loaders: dict[str, callable], path: Path, default: callable) -> callable:
+    file_extension = get_file_extension(path)
+    if file_extension in loaders:
```
diehlbw, line 25: style: loaders is just a normal dictionary correct?  Could this just be `loaders.get(file_extension, default)`

src/seismometer/data/loader/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        return loaders[file_extension]
+    return default
+
+
+def get_file_extension(path: Path) -> str:
```
diehlbw, line 30: Do you see this being used more commonly? I'm not sure I'd envision it being called outside get_loader_from_path so might not warrant its own function, vs directly in line~24.  I don't feel strongly about this one, if you envision reusability

src/seismometer/data/loader/__init__.py
```diff
@@ -23,14 +26,12 @@ def loader_factory(config: ConfigProvider, post_load_fn: ConfigFrameHook = None)
     SeismogramLoader
         Instance of the SeismogramLoader class, with load_data() method ready to be called
     """
-    prediction_loader: ConfigOnlyHook = prediction.parquet_loader
-    event_loader: ConfigOnlyHook = event.parquet_loader
 
     return SeismogramLoader(
         config,
-        prediction_fn=prediction_loader,
+        prediction_fn=prediction.loader,
```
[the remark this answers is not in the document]
eptgen: I agree. I kept the routing in the respective prediction and event files, but called these routing functions in the loader factory.

src/seismometer/data/loader/prediction.py
eptgen: Oops, I meant to get rid of that. Thanks for catching it.

src/seismometer/data/loader/utils.py
eptgen: I originally had utils because there were multiple functions that were shared between predictions and events. After making the changes you suggested, I determined there was no more need for one.

src/seismometer/data/loader/utils.py
eptgen: Inlined this function, I agree that it doesn't serve much use.

src/seismometer/data/loader/prediction.py
```diff
@@ -171,3 +231,35 @@ def _infer_datetime(dataframe, cols=None, override_categories=None):
...
+
+
+def _sv_loader(config: ConfigProvider, sep) -> pd.DataFrame:
+    """General loader for CSV or TSV files"""
+    if not config.features:  # no features ==> all features
+        dataframe = pd.read_csv(config.prediction_path, sep=sep)
+    else:
+        desired_columns = set(config.prediction_columns)
+
+        with warnings.catch_warnings():
```
diehlbw, line 243: Is there a specific FutureWarning you are suppressing that we can't address?  
It worries me a little to completely prevent visibility into these upcoming warnings, even when it's something like Future

diehlbw (changes requested):

src/seismometer/data/loader/prediction.py
eptgen: Good catch, the only reason I had that was because we already suppress that warning for Parquet files. I'm not sure why we suppress the warning there, but I don't think there's a need to do that for *SV files, so I got rid of it.

src/seismometer/data/loader/event.py
```diff
@@ -194,3 +258,28 @@ def _get_source_type(config: ConfigProvider, event: Event) -> str:
...
+        )
+
+        # since importing CSVs automatically cast numbers to ints, make sure the columns
+        # shared with predictions become strings so we don't have a type mismatch
+        defined_types = config.prediction_types
+        usage = config.usage
+        for col in [usage.entity_id, usage.context_id, usage.predict_time]:
+            if col is not None and defined_types[col] == "object":
+                events[col] = events[col].astype(str)
+    except BaseException:
```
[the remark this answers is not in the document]
eptgen: Changed!

diehlbw (approved):

gbowlin (approved):

PR #160, opened 2025-05-29
Add support for load-time data filtering and hierarchical cohorts

MahmoodEtedadi:
# Overview

Closes #xxx 

## Author Checklist

### DIFF
--- a/src/seismometer/configuration/model.py
+++ b/src/seismometer/configuration/model.py
@@ -2,7 +2,7 @@ import logging
 from pathlib import Path
 from typing import Any, List, Literal, Optional, Union
 
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, field_validator, model_validator
 
 logger = logging.getLogger("seismometer")
 
@@ -207,6 +207,70 @@ class BaseDetails(BaseModel):
     """ The group or groups to which the metric belongs."""
 
 
+class FilterRange(BaseModel):
+    """
+    A numeric range filter specifying minimum and/or maximum values.
+
+    Attributes
+    ----------
+    min : Optional[float]
+        The minimum allowed value (inclusive).
+    max : Optional[float]
+        The maximum allowed value (inclusive).
+    """
+
+    min: Optional[float] = None
+    max: Optional[float] = None
+
+
+class LoadTimeFilter(BaseModel):
+    """
+    A filter rule applied at data ingestion time to restrict the Seismogram data.
+
+    Attributes
+    ----------
+    source : str
+        The name of the column to filter.
+    action : Literal["include", "exclude", keep_top"]
+        The type of filtering to apply.
+        - "include": keep only rows matching the specified values or range
+        - "exclude": remove rows matching the specified values or range
+        - "keep_top": retain only the top MAXIMUM_NUM_COHORTS most frequent values (used with categorical variables)
+    values : Optional[List[Union[str, int, List[str], List[int]]]]
+        A list of allowed or disallowed values, by default None.
+    range : Optional[FilterRange]
+        A numeric range with min and/or max bounds, by default None. Typically used with numeric columns.
+    """
+
+    source: str
+    action: Literal["include", "exclude", "keep_top"] = "keep_top"
+    values: Optional[List[Union[str, int, List[str], List[int]]]] = None
+    range: Optional[FilterRange] = None
+
+    @model_validator(mode="after")
+    def validate_filter_config(self) -> "LoadTimeFilter":
+        """
+        Validates the consistency of filter fields after model is initialized.
+        Ensures appropriate combinations of action, values, and range.
+        """
+        if self.action in ("include", "exclude"):
+            if self.values is None and self.range is None:
+                raise ValueError(
+                    f"Filter with action '{self.action}' must specify at least one of 'values' or 'range'."
+                )
+            if self.values is not None and self.range is not None:
+                logger.warning(
+                    f"Filter on '{self.source}' specifies both 'values' and 'range'; "
+                    "'values' will be used and 'range' will be ignored."
+                )
+
+        elif self.action == "keep_top":
+            if self.values is not None or self.range is not None:
+                logger.warning(f"Filter on '{self.source}' with action 'keep_top' ignores 'values' and 'range'.")
+
+        return self
+
+
 class MetricDetails(BaseModel):
     """Contains details about a metric."""
 
@@ -382,7 +446,8 @@ class DataUsage(BaseModel):
     """
     metrics: list[Metric] = []
     """A list of all metrics to load."""
-
+    load_time_filters: list[LoadTimeFilter] = []
+    """A list of filters to apply at load time to reduce the working dataset."""
     censor_min_count: int = Field(10, ge=10)
     """ The minimum size of a cohort to be considered displayable. """
 
--- a/src/seismometer/seismogram.py
+++ b/src/seismometer/seismogram.py
@@ -124,6 +124,7 @@ class Seismogram(object, metaclass=Singleton):
         self._load_metadata()
 
         self.dataframe = self.dataloader.load_data(predictions, events)
+        self.dataframe = self._apply_load_time_filters(self.dataframe)
 
         self.create_cohorts()
         self._set_df_counts()
@@ -366,6 +367,37 @@ class Seismogram(object, metaclass=Singleton):
             self.thresholds = [0.8, 0.5]
 
         self.modelname: str = self._metadata.get("modelname", "UNDEFINED MODEL")
+
+    def _apply_load_time_filters(self, df: pd.DataFrame) -> pd.DataFrame:
+        """
+        Applies all load-time filters by building a single composite mask for filtering.
+        Returns a filtered DataFrame.
+        """
+        mask = np.ones(len(df), dtype=bool)
+        for rule in self.config.usage.load_time_filters:
+            column = rule.source
+            if column not in df.columns:
+                raise ValueError(f"Filter source column '{column}' not found in data.")
+            # Default: full True array (neutral for include/keep_top)
+            column_mask = np.ones(len(df), dtype=bool)
+            if rule.action == "keep_top":
+                top_values = df[column].value_counts().nlargest(MAXIMUM_NUM_COHORTS).index
+                column_mask = df[column].isin(top_values).values
+            elif rule.action in {"include", "exclude"}:
+                submask = np.ones(len(df), dtype=bool)
+                if rule.values:
+                    submask &= df[column].isin(rule.values).values
+                if rule.range:
+                    if rule.range.min is not None:
+                        submask &= df[column].values >= rule.range.min
+                    if rule.range.max is not None:
+                        submask &= df[column].values <= rule.range.max
+                column_mask = submask if rule.action == "include" else ~submask
+            else:
+                raise ValueError(f"Unsupported filter action: {rule.action}")
+            # Combine with global mask
+            mask &= column_mask
+        return df[mask]
 
     def create_cohorts(self) -> None:
         """Creates data columns for each cohort defined in configuration."""


src/seismometer/seismogram.py
```diff
@@ -436,6 +452,37 @@ def create_cohorts(self) -> None:
...
+                    raise ValueError(
+                        f"Cohort hierarchy '{hierarchy.name}' references undefined cohort source: '{level}'"
+                    )
+                resolved_levels.append(source_to_display[level])
+
+            resolved_hierarchies.append(CohortHierarchy(name=hierarchy.name, hierarchy=resolved_levels))
+
+        self.cohort_hierarchies = resolved_hierarchies
+
+    def _build_cohort_hierarchy_combinations(self):
```
gbowlin, line 475: why is this underscore prefixed when it is called by selection.py?

src/seismometer/seismogram.py
```diff
@@ -436,6 +452,37 @@ def create_cohorts(self) -> None:
             self.cohort_cols.append(disp_attr)
         logger.debug(f"Created cohorts: {', '.join(self.cohort_cols)}")
 
+    def _validate_and_resolve_cohort_hierarchies(self):
```
gbowlin, line 455: Is there a reason this validation can't live in `config.py`?

src/seismometer/configuration/model.py
```diff
@@ -456,6 +473,35 @@ def default_comparison(cls, comparison_time: str, values: dict) -> str:
...
+        list[CohortHierarchy]
+            The original list if valid.
+
+        Raises
+        ------
+        ValueError
+            If any column appears in more than one hierarchy.
+        """
+        all_fields = [col for h in hierarchies for col in h.hierarchy]
+        duplicates = {col for col in all_fields if all_fields.count(col) > 1}
```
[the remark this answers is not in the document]
MahmoodEtedadi: Updated to use `Counter`.

src/seismometer/controls/selection.py
```diff
@@ -140,15 +145,58 @@ def __init__(
...
+                        border="solid 1px var(--jp-border-color1)" if border else None,
+                        padding="var(--jp-cell-padding)" if border else None,
+                    ),
+                )
+            )
+
+        # Step 2: add non-hierarchical widgets
+        non_hierarchical_widgets = [self.selection_widgets[key] for key in options if key not in hierarchy_keys]
+        non_hierarchical_widget_box = Box(
+            children=non_hierarchical_widgets,
```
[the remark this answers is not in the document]
MahmoodEtedadi: The current structure is one hierarchy per row + all other cohorts in another line (as we had before).

src/seismometer/controls/selection.py
```diff
@@ -140,15 +145,58 @@ def __init__(
             selection_widget.observe(self._on_subselection_change, "value")
         self.value_update_in_progress = False
         self.title_box = HTML()
+
+        hierarchy_keys = set()  # all keys used in any hierarchy
+        hierarchy_widgets_list = []
+
+        # Step 1: group widgets by hierarchy
+        for hierarchy in self.hierarchies:
+            visible_keys = [key for key in hierarchy.hierarchy if key in self.selection_widgets]
```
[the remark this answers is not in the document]
MahmoodEtedadi: Changed data type to `Optional[list[CohortHierarchy]]`. Also, changed to use `options` instead of `self.selection_widgets`.
as far as the order in concerned, since we are checking this for each cohort list, it would probably be fine to leave it as is.

src/seismometer/controls/selection.py
```diff
@@ -140,15 +145,58 @@ def __init__(
...
+            ordered_groups = []  # groups of widgets corresponding to the hierarchy to display horizontally
+
+            if len(visible_keys) >= 2:
+                for i, key in enumerate(visible_keys):
+                    ordered_groups.append(self.selection_widgets[key])
+                    hierarchy_keys.add(key)
+                    if i < len(visible_keys) - 1:
+                        ordered_groups.append(HTML(value="→", layout=Layout(width="10px", align_self="flex-start")))
```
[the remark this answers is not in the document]
MahmoodEtedadi: Added `hierarch.name` as well.


image

src/seismometer/controls/selection.py
```diff
@@ -140,15 +145,58 @@ def __init__(
...
+        # Step 1: group widgets by hierarchy
+        for hierarchy in self.hierarchies:
+            visible_keys = [key for key in hierarchy.hierarchy if key in self.selection_widgets]
+            ordered_groups = []  # groups of widgets corresponding to the hierarchy to display horizontally
+
+            if len(visible_keys) >= 2:
+                for i, key in enumerate(visible_keys):
+                    ordered_groups.append(self.selection_widgets[key])
+                    hierarchy_keys.add(key)
+                    if i < len(visible_keys) - 1:
```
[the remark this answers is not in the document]
MahmoodEtedadi: Updated.

src/seismometer/seismogram.py
```diff
@@ -387,11 +392,22 @@ def _apply_load_time_filters(self, df: pd.DataFrame) -> pd.DataFrame:
...
+                    except (TypeError, ValueError) as e:
+                        bounds_str = ", ".join(
+                            f"{label}={val}"
+                            for label, val in [("min", rule.range.min), ("max", rule.range.max)]
+                            if val is not None
+                        )
+                        raise ValueError(
+                            f"Range filter on column '{column}' failed. Ensure values are comparable to {bounds_str}."
+                            f" Underlying error: {e}"
+                        )
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

src/seismometer/seismogram.py
MahmoodEtedadi: Since the check is across different config attributes (use cohorts display names to update hierarchy lists), I thought it would be more appropriate in `seismogram.py`.

src/seismometer/seismogram.py
MahmoodEtedadi: Now it is only called in `seismogram.py`.

src/seismometer/configuration/model.py
```diff
@@ -207,6 +207,85 @@ class BaseDetails(BaseModel):
     """ The group or groups to which the metric belongs."""
 
 
+class FilterRange(BaseModel):
```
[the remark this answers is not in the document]
MahmoodEtedadi: Do not need to update pngs as we do not show cohort selection widget.
Updated `configuration.rst` and user guide.

src/seismometer/controls/selection.py
```diff
@@ -140,15 +145,58 @@ def __init__(
...
+        # Step 2: add non-hierarchical widgets
+        non_hierarchical_widgets = [self.selection_widgets[key] for key in options if key not in hierarchy_keys]
+        non_hierarchical_widget_box = Box(
+            children=non_hierarchical_widgets,
+            layout=Layout(
+                display="flex",
+                flex_flow="row wrap",
+                align_items="flex-start",
+                grid_gap="20px",
+                border="solid 1px var(--jp-border-color1)" if border else None,
```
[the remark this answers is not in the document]
MahmoodEtedadi: I think other widgets already call this class and am not sure if would gain much by defining this in style. This is specific to the visualization for hierarchies (with -> in between).

src/seismometer/controls/selection.py
```diff
@@ -140,15 +150,64 @@ def __init__(
...
+                        display="flex",
+                        flex_flow="row wrap",
+                        align_items="flex-start",
+                        grid_gap="3px",
+                        border="solid 1px var(--jp-border-color1)" if border else None,
+                        padding="var(--jp-cell-padding)" if border else None,
+                    ),
+                )
+            )
+
```
gbowlin, line 183: Should we make a Hierarchy  control that does this grouping automatically? Then we can create the right control class based on group content. Right now it feels odd that you have to choose to add the arrows and `_on_subselection_change` content.

src/seismometer/seismogram.py
```diff
@@ -367,6 +371,49 @@ def _load_metadata(self):
 
         self.modelname: str = self._metadata.get("modelname", "UNDEFINED MODEL")
 
+    def _apply_load_time_filters(self, df: pd.DataFrame) -> pd.DataFrame:
```
gbowlin, line 374: Is it possible to have the loadtime config just create a FilterRule (we can add a filter rule for TopK? 

Then we can just build the FilteRules from rules. I think this would make it easier to 1. Create a bunch of rules, 2. apply conjuction. 

It also means you can separate out rule logic from dataframe masking.

src/seismometer/table/fairness.py
```diff
@@ -398,11 +400,17 @@ def __init__(
...
+            List of cohort hierarchies to consider, by default None.
+        hierarchy_combinations: Optional[dict[tuple[str], pd.DataFrame]], optional
+            Mapping of each hierarchy to valid combinations of values across its levels, by default None.
         """
         self.model_options_widget = model_options_widget
         default_metrics = default_metrics or metric_names
         self.metric_list = MultiselectDropdownWidget(metric_names, value=default_metrics, title="Fairness Metrics")
-        self.cohort_list = MultiSelectionListWidget(cohort_dict, title="Cohorts")
+        self.cohort_list = MultiSelectionListWidget(
+            cohort_dict, title="Cohorts", hierarchies=hierarchies, hierarchy_combinations=hierarchy_combinations
```
gbowlin, line 412: Whats the use case / user story for including a hierarchy in the fairness table? In the fairness widget, Cohorts are not filters, but a set of "categories"  to view fairness across.  This makes the hierarchy control unexpected to me.  

image

If we do want that to be a filter, then this feels like we should move the hierarchies out of Cohort and have something different like a Filter list and a Cohort list.  (which starts getting really busy)

src/seismometer/seismogram.py
MahmoodEtedadi: Added `topk`, `nottopk` rule and here we just apply filters from `FilterRule`.

src/seismometer/controls/selection.py
```diff
@@ -140,15 +150,64 @@ def __init__(
...
+        # Step 2: add non-hierarchical widgets
+        non_hierarchical_keys = [key for key in options if key not in hierarchy_keys]
+        # Add spacer label only if there are cohort hierarchies
+        non_hierarchical_widgets = (
+            [Label(value="", layout=Layout(min_width="120px", align_self="flex-start"))] if self.hierarchies else []
```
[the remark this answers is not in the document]
MahmoodEtedadi: Removed the empty title.

src/seismometer/controls/selection.py
MahmoodEtedadi: Created a new HierarchicalSelectionWidget class to handle hierarchy widgets.

src/seismometer/controls/selection.py
```diff
@@ -169,18 +228,58 @@ def disabled(self, disabled: bool):
...
+        """Sets the observable value and updates options based on parent selections."""
         if self.value_update_in_progress:
             return
-        if change and change["owner"] in self.selection_widgets.values():
-            self.value = {k: tuple(v.value) for k, v in self.selection_widgets.items() if len(v.value)}
+
+        # Track latest selections for chaining
+        selected = {k: tuple(v.value) for k, v in self.selection_widgets.items() if len(v.value)}
+
+        for hierarchy in self.hierarchies:
```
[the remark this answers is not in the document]
MahmoodEtedadi: I think we considered disabling options, but that would mean not selecting any options means empty selection but throughout we treat this as selecting all (or all available).

src/seismometer/controls/selection.py
```diff
@@ -91,6 +93,115 @@ def get_selection_text(self) -> str:
...
+    Widget for displaying and chaining a hierarchy of selection widgets.
+    """
+
+    value = traitlets.Dict(help="Current selection across the hierarchy levels")
+
+    def __init__(
+        self,
+        hierarchy: CohortHierarchy,
+        widgets: dict[str, ValueWidget],
+        combinations: Optional["pd.DataFrame"] = None,
```
gbowlin, line 107: Should this be optional or required?  I feel like it should be required. If its not required, then there are no restrictions between elements right?

I was thinking a different data structure for combinations might make sense, but maybe not. pd.Dataframe is generic, which means we don't get great typing. If we did a list of valid tuples, then we end up having to do a lot of search reductions (based on the types already selected) which you already implmeneted as dataframe. 

In any case, please don't pass in widgets, create your own so that we don't have to manage them this way.

src/seismometer/controls/selection.py
```diff
@@ -132,23 +249,56 @@ def __init__(
...
+        # Step 1: group widgets by hierarchy
+        for hierarchy in self.hierarchies:
+            visible_keys = [key for key in hierarchy.column_order if key in options]
+            hierarchy_keys.update(visible_keys)
+            widget_group = HierarchicalSelectionWidget(
+                hierarchy=hierarchy,
+                widgets={k: self.selection_widgets[k] for k in visible_keys},
+                combinations=self.hierarchy_combinations.get(tuple(hierarchy.column_order)),
+                border=border,
+            )
```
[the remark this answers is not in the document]
MahmoodEtedadi: Moved widget creation to inside the class.

src/seismometer/seismogram.py
```diff
@@ -367,6 +372,57 @@ def _load_metadata(self):
...
+                            submask = range_rule.mask(df)
+                    except (TypeError, ValueError) as e:
+                        bounds_str = ", ".join(
+                            f"{label}={val}"
+                            for label, val in [("min", rule.range.min), ("max", rule.range.max)]
+                            if val is not None
+                        )
+                        raise ValueError(
+                            f"Range filter on column '{column}' failed. Ensure values are comparable to {bounds_str}."
+                        ) from e
```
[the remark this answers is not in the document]
MahmoodEtedadi: FINC.

src/seismometer/data/filter.py
```diff
@@ -215,6 +215,11 @@ def mask(self, data: pd.DataFrame) -> pd.Index:
             Index masked to only the rows that match the FilterRule.
         """
         relation = FilterRule.method_router[self.relation]
+        if self.relation in ["<=", ">=", "<", ">"]:
+            try:
+                return relation(data, self.left, self.right)
+            except (TypeError, ValueError) as e:
+                raise ValueError(f"Values in '{self.left}' must be comparable to '{self.right}'.") from e
```
gbowlin, line 222: Should this try be in  mask the method router lambda

We could replace `lambda x, y, z: x[y] <= z,` with a `def rows_with_colum_compared_to(df, col_name, comparison, value) -> mask` so that we give an example of not require a lambda in the method router.

src/seismometer/seismogram.py
```diff
@@ -380,33 +380,31 @@ def _apply_load_time_filters(self, df: pd.DataFrame) -> pd.DataFrame:
         if not self.config.usage.load_time_filters:
             return df
 
-        mask = np.ones(len(df), dtype=bool)
+        combined_rule = FilterRule.all()
 
         for rule in self.config.usage.load_time_filters:
             column = rule.source
             if column not in df.columns:
                 raise ValueError(f"Filter source column '{column}' not found in data.")
```
gbowlin, line 388: This value error can also move to `filter.py`. 

In fact do we have a good reason not to put most of this logic into `FilterRule.from_filter_config(filter_list:list[LoadTimeFilter])` and maybe rename `LoadTimeFilter` to `FilterConfig`

src/seismometer/data/filter.py
```diff
@@ -387,6 +411,48 @@ def from_cohort_dictionary(cls, cohort_dict: dict[str, tuple[any]] | None = None
...
+
+            if rule.values:
+                subrule = cls.isin(column, rule.values)
+
+            elif rule.range:
+                if rule.range.min is not None or rule.range.max is not None:
+                    if rule.range.min is not None:
+                        subrule = cls.geq(column, rule.range.min)
+                        if rule.range.max is not None:
+                            subrule &= cls.leq(column, rule.range.max)
```
diehlbw, line 434: Commented in the model definition, but I think the half-open and exclusive max is more intuitive for sets of adjacent filters if the data is continuous.

src/seismometer/data/filter.py
```diff
@@ -24,6 +24,8 @@ class FilterRule(object):
         "notna": lambda x, y, z: ~x[y].isna(),
         "isin": lambda x, y, z: x[y].isin(z),
         "notin": lambda x, y, z: ~x[y].isin(z),
+        "topk": lambda x, y, z: x[y].isin(x[y].value_counts().nlargest(z).index),
+        "nottopk": lambda x, y, z: ~x[y].isin(x[y].value_counts().nlargest(z).index),
```
diehlbw, line 28: Does this handle ties in the way we want?
If ties span across k, I _think_ this ends up being reliant on whatever value_counts default sort does.  

If all values have uniform frequency we get rows of the first z values

MahmoodEtedadi: Changed it to `.value_counts(ascending=False).sort_index()` to consistently handle ties.

diehlbw: what is the desired behavior on ties?

In a hypothetical uniform case of values {a,b,c,d,e,f,g,h,i,j} each being used 10 times, what should top5 include?

src/seismometer/data/filter.py
```diff
@@ -411,6 +471,48 @@ def from_cohort_dictionary(cls, cohort_dict: dict[str, tuple[any]] | None = None
                 rule = rule & cls.isin(key, cohort_dict[key])
         return rule
 
+    @classmethod
+    def from_config(cls, rule: "FilterConfig") -> "FilterRule":
+        """Creates a FilterRule from a high-level FilterConfig specification."""
+        from seismometer.seismogram import MAXIMUM_NUM_COHORTS
```
gbowlin, line 477: `MAXIMUM_NUM_COHORTS` should just be a class level property that we can set on the FilteRule class like with

`MIN_ROWS: Optional[int] = 10`

Then we can override it based on the `MAX_NUM_COHORTS` as needed

importantly, there should be no reason to import seismometer.seismogram into this class. 

I would even recommend moving the model for FilterConfig into this class and just importing that model into `model.py` so that this file remains self-conftained.

src/seismometer/data/filter.py
```diff
@@ -411,6 +471,48 @@ def from_cohort_dictionary(cls, cohort_dict: dict[str, tuple[any]] | None = None
...
+
+        column = rule.source
+        if rule.action == "keep_top":
```
gbowlin, line 480: If we are going to have `keep_top`, shouldn't we store the value for how many to keep?


    - source: loc 
      action: keep_top
      count: 4

If count isn't set, then we can use MAXIMUM_NUM_COHORTS as a default.

MahmoodEtedadi: Added `count` to `FilterConfig` data class.

src/seismometer/data/filter.py
```diff
@@ -24,14 +89,14 @@ class FilterRule(object):
         "notna": lambda x, y, z: ~x[y].isna(),
         "isin": lambda x, y, z: x[y].isin(z),
         "notin": lambda x, y, z: ~x[y].isin(z),
-        "topk": lambda x, y, z: x[y].isin(x[y].value_counts().nlargest(z).index),
-        "nottopk": lambda x, y, z: ~x[y].isin(x[y].value_counts().nlargest(z).index),
+        "topk": partial(apply_topk_filter, op="topk"),
+        "nottopk": partial(apply_topk_filter, op="nottopk"),
```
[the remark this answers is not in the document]
MahmoodEtedadi: Updated as suggested.

gbowlin (dismissed):

src/seismometer/configuration/model.py
```diff
@@ -207,6 +208,91 @@ class BaseDetails(BaseModel):
...
+    """Exclusive upper bound for allowed values."""
+
+
+class FilterConfig(BaseModel):
+    """A configurable rule for filtering data based on values in a specific column."""
+
+    source: str
+    """The name of the column to filter."""
+
+    action: Literal["include", "exclude", "keep_top"] = "keep_top"
```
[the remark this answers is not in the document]
MahmoodEtedadi: Changed the default value for `action` to `None` and infer the value based on other attributes:
1. If `count` provided -> `keep-top`
2. elif `values` or `range` provided -> `include`
3. else, raise an error

src/seismometer/seismogram.py
```diff
@@ -367,6 +372,20 @@ def _load_metadata(self):
...
+        """
+        if not self.config.usage.load_time_filters:
+            return df
+
+        for rule in self.config.usage.load_time_filters:
+            if rule.source not in df.columns:
+                raise ValueError(f"Filter source column '{rule.source}' not found in data.")
+        FilterRule.MAXIMUM_NUM_COHORTS = MAXIMUM_NUM_COHORTS
+        return FilterRule.from_filter_config_list(self.config.usage.load_time_filters).filter(df)
+
```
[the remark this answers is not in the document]
MahmoodEtedadi: As we talked about it, it might be more beneficial to have cohort hierarchies separate from cohorts and not enforce the 25 limit on the number of different categories they can have. 

We will add this in a subsequent PR.

src/seismometer/configuration/model.py
```diff
@@ -207,6 +208,107 @@ class BaseDetails(BaseModel):
...
+            elif self.values is not None or self.range is not None:
+                self.action = "include"
+                logger.info(f"Inferred action 'include' for filter on '{self.source}'")
+            else:
+                raise ValueError(
+                    f"Filter on '{self.source}' must specify one of 'values', 'range', or 'count', "
+                    "or explicitly set an action."
+                )
+
+        if self.action in ("include", "exclude"):
```
diehlbw, line 266: To be comprehensive about the clearing out unused config  
Should this block set count to None?

MahmoodEtedadi: FINC.

PR #166, opened 2025-06-12
Add OpenTelemetry support into Seismometer

adamhutchings:
# Overview
This PR adds support for emitting OpenTelemetry-style metrics from many of the widgets and plot API functions.

## Description of changes
- New `OpenTelemetryRecorder` class which outputs metrics. The actual output locations (for now, console / file dump + to Prometheus) are configurable.
- Calls to `OpenTelemetryRecorder.populate_metrics` are added to most of the widgets.

## Work to be done before this undrafted
- Clean up places where metrics are populated (e.g. make it more consistent across some of the plots)
- Do docstrings and comments
- Do docs
- Make sure tests are passing

## Author Checklist


adamhutchings:
Some comments on the current state of the PR:
- By the asynchronous nature of OpenTelemetry metric exporting, testing of the code has been fairly manual, using the diabetes dataset. Suggestions for testing are welcome.
- There is a utility for recording data for combinations of cohorts, otel.log_by_cohort, which is yet unused in the code itself. During the code review process or further development, I will use it to clean up some of the existing metric-logging code (although not all of the patterns are amenable to such use). For automation support and configuration of which such sets of cohorts to select and log metrics on, I will be specifying some further configuration in a later RFC.

docs/integration_guide/backend.rst
```diff
@@ -0,0 +1,126 @@
...
+      # For communication with a backend
+      networks:
+        - metricsnetwork
+
+  networks:
+    metricsnetwork:
+      external: true
+
+Before starting either Docker container, run the command
+``docker create network metricsnetwork``. This will actually make the
```
gbowlin, line 115: I think this is ``docker network create metricsnetwork`

```suggestion
``docker network create metricsnetwork``. This will actually make the
```

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,352 @@
...
+    """Get from the environment whether we should export metrics at all, or not.
+
+    Returns
+    -------
+    bool
+        Whether or not all OTel outputs will be disabled.
+    """
+    raw_stop = os.getenv("SEISMO_NO_OTEL", "FALSE")
+    if raw_stop not in ["TRUE", "FALSE"]:
+        logger.warn("Unrecognized value for SISMO_NO_OTEL. Defaulting to false (metrics will be output ...)")
```
firste, line 33:
```suggestion
        logger.warn("Unrecognized value for SEISMO_NO_OTEL. Defaulting to false (metrics will be output ...)")
```

src/seismometer/report/profiling.py
```diff
@@ -189,6 +196,23 @@ def __init__(
...
             self._deserialize_alerts()
 
+        alert_types = ["imbalance", "missing", "zeros"]
+        # TODO: add units of percent for each of these
+        self.recorder = otel.OpenTelemetryRecorder(metric_names=alert_types, name="ydata profiling report")
+        for alert in self._parsed_alerts.alerts:
+            # alert.display_html is something like:
+            # <a href="#pp_var_4085253654425318375"><code>A1Cresult</code></a> is highly imbalanced (54.4%).
+            # So we're going to extract the relevant parts for logging: A1Cresult and 54.4%.
+            results = re.search(r"<code>(.+)</code>.+\((.*)%\)", alert.display_html)
```
firste, line 206: Are you sure you need a regex for this? It looks like ydata profiling does have a way to programmatically get at the alerts.

.devcontainer/docker-compose.yml
```diff
@@ -22,3 +21,11 @@ services:
...
     # Overrides default command so things don't shut down after the process ends.
     command: sleep infinity
+
+    # For communication with a backend
+    networks:
+      - metricsnetwork
+
+networks:
+  metricsnetwork:
+    external: true
```
firste, line 31: My dev container no longer builds with this PR. Flipping this to false makes it spin up. Is there a different approach that you can take to avoid the errors?

setup.cfg
```diff
@@ -36,6 +36,8 @@ install_requires =
     pydantic>=2.6.3,<3
     ydata-profiling>=4.8.3,<5
     great-tables>=0.11.0,<1
+    opentelemetry-sdk>=1.34.1
+    opentelemetry-exporter-otlp>=1.34.1
```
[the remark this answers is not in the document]
gbowlin: See line 46 below
```toml
[options.extras_require]
all =
    %(dev)s
    %(docs)s
```

We could add a `%(otel)s` for these packages.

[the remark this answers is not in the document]
adamhutchings: They are now optional.

src/seismometer/seismogram.py
```diff
@@ -439,4 +459,110 @@ def get_ordinal_categorical_groups(self, max_cat_size):
...
+            "predict_time": sg.predict_time,
+            "time_zero": sg.time_zero,
+            "thresholds": sg.thresholds,
+            "entity_keys": sg.entity_keys,
+            "outcome": sg.outcome,
+            "intervention": sg.intervention,
+            "comparison_time": sg.comparison_time,
+            "available_cohort_groups": sg.available_cohort_groups,
+            "config.entity_id": sg.config.entity_id,
+        }
```
gbowlin, line 492: Do we actually need to call this? some of these are just defaults, and many could change on reload or with a new dataset right?

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+            fn = show_cohort_summaries
+        case "target_feature_summary":
+            fn = target_feature_summary
+        case _:
+            raise ValueError(f"Unknown function name: {function_name}")
+    fn = allowed_export_names[function_name]
+    fn(*args, **kwargs)
+
+
+def extract_arguments(argument_names: list[str], run_settings: dict) -> dict:
```
gbowlin, line 543: This feels like pydantic data models should be able to load an appropriately rich object itself, which would avoid needing to do this separately.

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+    Returns
+    -------
+    dict
+        The options and values we found in options.
+    """
+    return (
+        {arg: run_settings["options"][arg] for arg in argument_names if arg in run_settings["options"]}
+        if "options" in run_settings
+        else {}
+    )
```
gbowlin, line 567: While valid python, it reads better with if blocks:

```suggestion
        if "options" in run_settings:
            return {arg: run_settings["options"][arg] for arg in argument_names if arg in run_settings["options"]}
        return {}
```

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
+import functools
```
gbowlin, line 1: In general, can we break this file into smaller bits by functionality? 

We have YAML Config,  Otel Communctions (socket, Manager that uses the socket), and Metric Population APIs.

Feels like 

1. seismometer.config for YAML/data model objects. 
2. seismometer.core.otel for Otel communication classes
3. seismometer.data.metrics for metric apis (that need to work well with data apis)

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+try:
+    from opentelemetry.exporter.otlp.proto.grpc.metric_exporter import OTLPMetricExporter
+    from opentelemetry.metrics import Histogram, Meter, UpDownCounter, set_meter_provider
+    from opentelemetry.sdk.metrics import MeterProvider
+    from opentelemetry.sdk.metrics.export import ConsoleMetricExporter, PeriodicExportingMetricReader
+    from opentelemetry.sdk.resources import SERVICE_NAME, Resource
+
+    TELEMETRY = True
+except ImportError as e:
+    logger.warning(f"Was not able to use metric exporting (message {str(e)}). Not using telemetry ...")
```
gbowlin, line 30: if telemetry is disabled, you don't need to warn.

src/seismometer/configuration/config.py
```diff
@@ -100,6 +102,13 @@ def _resolve_other_paths(
         """
         # Coerce to Path or override with Path
         self.config.usage_config = Path(usage_config) if usage_config is not None else Path(self.config.usage_config)
+        self.config.automation_config = (
```
gbowlin, line 105: You should move your otel configuration here so that we can read a full set of config files before even importing the OTEL stuff, that way we can easily turn otel off if its not configured.

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+        Whether or not all OTel outputs will be disabled.
+    """
+    raw_stop = os.getenv("SEISMO_NO_OTEL", "FALSE")
+    if raw_stop not in ["TRUE", "FALSE"]:
+        logger.warn("Unrecognized value for SEISMO_NO_OTEL. Defaulting to false (metrics will be output ...)")
+        raw_stop = "FALSE"
+    return raw_stop == "TRUE"
+
+
+STOP_ALL_OTEL = config_otel_stoppage()
```
gbowlin, line 49: Do we need to do this during import? can we just leave otel off until we turn it on?

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+        return yaml.safe_load(file)["otel_info"]
+    except FileNotFoundError:
+        raise Exception("Could not find usage config file for metric setup!")
+    except (KeyError, TypeError):
+        logger.warning(f"No OTel config found in {file_path}. Will be defaulting ...")
+        return {}
+
+
+# This will be set once usage_config.yml is done downloading, in run_startup.
+OTEL_INFO = {}
```
gbowlin, line 81: Do we need to use global maps? Is there a config object was can pass instead?

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+                logger.warning("Connecting to port failed. Ignoring ...")
+                if otel_collector_reader is not None:
+                    otel_collector_reader.shutdown()
+            else:
+                self.readers.append(otel_collector_reader)
+        self.resource = Resource.create(attributes={SERVICE_NAME: "Seismometer"})
+        self.meter_provider = MeterProvider(resource=self.resource, metric_readers=self.readers)
+        set_meter_provider(self.meter_provider)
+
+    def __del__(self):
```
gbowlin, line 186: We probably want stop and start methods that someone else can call in case we wan to pause exports for a while.

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+            # More complex: go through each combination of attributes.
+        else:
+            if cohorts:
+                keys = cohorts.keys()
+                # So if column "A" has attributes A_false and A_true, and B has 1, 2, 3, then
+                # we will exhaust through all combinations of these attributes.
+                selections = list(itertools.product(*[cohorts[key] for key in keys]))
+                # Now we put them into a dictionary for ease of processing
+                selections = [dict(zip(keys, s)) for s in selections]
+            else:
```
gbowlin, line 352: For such a big if else, you may want to create two different functions to map to, if for nothing else than to make it easer to test the two cases separately.

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+            The attributes we want to associate to all of the logs from this.
+        """
+        for metric in df.columns:
+
+            def maker(frame):
+                return frame.set_index(col_name)[[metric]].to_dict()
+
+            log_all = get_metric_config(metric)["log_all"]
+            if log_all:
+                self.log_by_cohort(base_attributes=base_attributes, dataframe=df, cohorts=cohorts, metric_maker=maker)
```
gbowlin, line 393: if not log all, what do we do, log none?

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+            pass
+
+    get_metric_creator = noop  # noqa: F811
+
+
+# For debug purposes: dump to stdout, and also to exporter path
+export_manager = ExportManager(file_output_path=sys.stdout, export_port=4317)
+
+
+def ready_for_serialization(obj):
```
gbowlin, line 427: do you need recursive conversion or could you use a pydantic data model instead? If we need custom serialization, I'd look at moving what you can to seisometer.core.io

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+    access to the Jupyter frontend.
+    """
+    with open("metric-automation.yml", "w") as automation_file:
+        sg = Seismogram()
+        call_history = ready_for_serialization(sg._call_history)
+        yaml.dump(call_history, automation_file)
+
+
+# Writing a full list here for security reasons.
+allowed_export_names = {
```
gbowlin, line 471: Is there a way to avoid having a list and instead have a collection of callable?  We have @export already, so maybe we can expect to import the list of names from the api module's so that we don't need to do this here?

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+        _plot_leadtime_enc,
+        model_evaluation,
+        plot_binary_classifier_metrics,
+        plot_cohort_evaluation,
+    )
+    from seismometer.api.reports import feature_alerts, feature_summary  # target_feature_summary
+
+    # from seismometer.api.templates import show_cohort_summaries
+
+    match function_name:
```
gbowlin, line 597: This match statement makes me wonder if we want to extract some of the callables so that we can automate variable retrieval from configuration and then calling them like we do in the `explore.py`'s `_try_generate_plot()`

src/seismometer/seismogram.py
```diff
@@ -97,6 +104,31 @@ def _initialize_attrs(self):
...
+        """
+
+        # Some of the plot functions take data frames as arguments, and we do not want to cache those.
+        # So, we will look through args and kwargs and replaec any arguments of DataFrame type.
+        def replace_df_map(val):
+            return "[Data Frame]" if isinstance(val, pd.DataFrame) else val
+
+        args = map(replace_df_map, args)
+        kwargs = {k: replace_df_map(kwargs[k]) for k in kwargs}
+        self._call_history[fn_name].append({"args": args, "kwargs": kwargs, "extra_info": extra_info(args, kwargs)})
```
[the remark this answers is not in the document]
adamhutchings: I did it this way because while removing key-value pairs in `kwargs` would keep it relatively undisturbed, dropping elements in the ordered list of `args` may result in some trouble when we try to call the function with the stored parameters (we could do some trickery to insert the dataframe back in the right place, but that's probably more trouble than keeping the list the same length but replacing dataframes with a dummy value).

src/seismometer/api/plots.py
```diff
@@ -77,6 +73,7 @@ def _plot_cohort_hist(
     cohort_col: str,
     subgroups: list[str],
     censor_threshold: int = 10,
+    filter_zero_one: bool = False,
```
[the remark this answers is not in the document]
adamhutchings: It looks like it is. I'll fix it.

src/seismometer/api/plots.py
```diff
@@ -419,6 +476,19 @@ def _plot_cohort_evaluation(
...
+        cohort_categories = list(set(plot_data["cohort"]))
+        recorder.log_by_cohort(
+            base_attributes=base_attributes,
+            dataframe=plot_data,
+            cohorts={"cohort": cohort_categories, "Threshold": [t * 100 for t in thresholds]},
+            intersecting=True,
+        )
+        recorder.log_by_column(
+            df=plot_data, col_name="Threshold", cohorts={"cohort": cohort_categories}, base_attributes=base_attributes
+        )
```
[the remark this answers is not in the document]
adamhutchings: log_by_column will log every row of the dataframe's attributes which we have log_all set on for, whereas the first call will only log from the list of thresholds.

src/seismometer/data/otel.py
adamhutchings: I added an option to log some if needed.

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,694 @@
...
+import logging
+import operator
+import os
+import socket
+import sys
+from typing import Any, Callable, Dict, List
+
+import numpy as np
+import pandas as pd
+import yaml
```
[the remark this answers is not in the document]
adamhutchings: This comment will be addressed above -- closing it for now to avoid duplicates.

src/seismometer/data/otel.py
adamhutchings: It has now been broken up.

src/seismometer/data/otel.py
adamhutchings: I was thinking it might be effective to associate each metric-exporting function with a specification of how it should be read and automated. At least, there is currently a lot of repetitive logic, so I think I could clean most of it up.

src/seismometer/data/otel.py
adamhutchings: This has been done.

src/seismometer/data/otel.py
adamhutchings: This has been done.

src/seismometer/data/otel.py
adamhutchings: Good point.

src/seismometer/data/otel.py
adamhutchings: This is true.

src/seismometer/seismogram.py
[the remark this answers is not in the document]
adamhutchings: I moved the methods.

src/seismometer/table/analytics_table.py
```diff
@@ -330,6 +331,31 @@ def analytics_table(self) -> HTML:
...
+                    if column.startswith(f"{value}_"):
+                        metric_name = column.lstrip(f"{value}_")
+                        base_attributes = {"metric value": value}
+                        break
+            if metric_name is None:
+                continue
+            # The table in the graphic is definitely indexed by score + target, so
+            # we will be storing those as attributes and the rest of the row as a
+            # bunch of key-value pairs.
+            metric_name = metric_name.replace("\xa0", " ")
```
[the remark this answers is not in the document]
adamhutchings: It has been moved.

src/seismometer/data/metric_apis.py
```diff
@@ -0,0 +1,252 @@
...
+            self.metric_names = []
+
+        def populate_metrics(self, *args, **kwargs):  # noqa: F811
+            pass
+
+        def log_by_cohort(self, *args, **kwargs):  # noqa: F811
+            pass
+
+        def log_by_column(self, *args, **kwargs):  # noqa: F811
+            pass
```
gbowlin, line 252: If you want to go this path, reather than overwrite the recorder use a class cluster and `__new__` to provide the right variant:

```python
class OpenTelemetryRecorder:
    def __new__(cls, *args, **kwargs):
        try:
            # Try to import the otel module
           from opentelemetry.metrics import Histogram, UpDownCounter
            # If successful, return an instance of the real class
            return super().__new__(RealRecorder)
        except ImportError:
            # If not, return an instance of the no-op fallback
            return super().__new__(NoOpRecorder)

    def populate_metrics(self, *args, **kwargs):
        raise NotImplementedError("Subclass must implement this method.")


class RealRecorder(OpenTelemetryRecorder):
    def __init__(self, *args, **kwargs):
        print("Initialized NoOpRecorder")

    def do_something(self):
        print("Doing real work!")


class NoOpRecorder(OpenTelemetryRecorder):
    def __init__(self, *args, **kwargs):
        print("Initialized NoOpRecorder")

    def do_something(self):
        pass  # intentionally does nothing
```

Just feels like a nicer way to accomplish the same thing. 

This also means you can alert the user who tries to turn on or off telemetry in a notebook that no matter what they do they cannot turn on telemetry if otel is not present.

gbowlin: Same applies to the otel ExportManager class

src/seismometer/api/plots.py
```diff
@@ -111,7 +113,34 @@ def _plot_cohort_hist(
...
+
+            cohorts = {col: df_other_attributes[col] for col in df_other_attributes.columns}
+            base_attributes = ts_attributes | {cohort_col: subgroup}
+            recorder.log_by_cohort(
+                base_attributes=base_attributes,
+                dataframe=bin_series,
+                cohorts=cohorts,
+                intersecting=False,
+                metric_maker=maker,
+            )
```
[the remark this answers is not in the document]
gbowlin: Yeah, thats not good as the bins can move around or just not mean anything reasonable, if we are going to log this type of data, quantiles are much better to log. Maybe don't make this plot log anything.

src/seismometer/data/metric_apis.py
```diff
@@ -0,0 +1,252 @@
...
+                self._log_to_instrument(attributes | {"key": k}, instrument, v)
+        elif isinstance(data, str):
+            raise Exception(f"Cannot log strings as metrics in OTel. Tried to log {data} as a metric.")
+        else:
+            raise Exception(f"Unrecognized data format for OTel logging: {str(type(data))}")
+
+    def log_by_cohort(
+        self,
+        base_attributes: Dict[str, Any],
+        dataframe: pd.DataFrame,
```
gbowlin, line 142: put dataframe first as thats our general convention.

src/seismometer/api/plots.py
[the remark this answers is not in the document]
adamhutchings: The plot no longer logs anything.

src/seismometer/__init__.py
```diff
@@ -65,3 +67,6 @@ def run_startup(
     sg = Seismogram(config, loader)
 
     sg.load_data(predictions=predictions_frame, events=events_frame)
+
+    otel.OTEL_INFO = otel.read_otel_info(config.config.usage_config)
+    sg.load_automation_config(config.config.automation_config)
```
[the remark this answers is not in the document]
adamhutchings: Yes! It has been done.

src/seismometer/api/plots.py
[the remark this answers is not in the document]
gbowlin: can you make it one method `log_by_cohort(stuff, required_labels = [ {"Threshold": t * 100} for t in thresholds])`

src/seismometer/data/metric_apis.py
adamhutchings: This has been done.

src/seismometer/data/metric_apis.py
adamhutchings: This has been done.

src/seismometer/table/analytics_table.py
```diff
@@ -330,6 +331,31 @@ def analytics_table(self) -> HTML:
...
+        for column in data.columns:
+            base_attributes = {}
+            metric_name = None
+            if column in METRICS:
+                metric_name = column
+            else:
+                for value in self.metric_values:
+                    if column.startswith(f"{value}_"):
+                        metric_name = column.lstrip(f"{value}_")
+                        base_attributes = {"metric value": value}
```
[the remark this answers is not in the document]
adamhutchings: Done!

src/seismometer/api/plots.py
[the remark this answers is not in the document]
gbowlin: We are still doing something similar in `_model_evaluation`

src/seismometer/table/analytics_table.py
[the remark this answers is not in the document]
gbowlin: Okay, now that i see the rest of the code around it, it feels like this is still doing something off.. I thought we were doing standard pandas column handling for seismometer where we have suffixes of type `_Value` or `_Time`

This seems specific to the analytics table itself, so a function in the analytics table class itself seems more appropriate to undo its own mangling.

src/seismometer/data/otel.py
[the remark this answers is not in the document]
gbowlin: This is still valid because autometics now calls the load function for OTEL config, As the only caller, it doesnt make sense to have the function in otel.py

src/seismometer/core/autometrics.py
```diff
@@ -0,0 +1,282 @@
...
+    """Read all metric exporting and automation info.
+
+    Parameters
+    ----------
+    config : OtherInfo
+        The configuration object handed in during Seismogram initialization.
+    """
+    global OTEL_INFO, STOP_ALL_OTEL
+    OTEL_INFO = read_otel_info(config.usage_config)
+    Seismogram().load_automation_config(config.automation_config)
```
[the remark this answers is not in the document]
gbowlin: I think this is still relevant overall, as we should think about how a data scientist will interact with this via notebook and command line/script.

setup.py
```diff
@@ -7,4 +7,13 @@
 
 if __name__ == "__main__":
     site.ENABLE_USER_SITE = "--user" in sys.argv[1:]
-    setuptools.setup()
+    setuptools.setup(
+        name="runmetrics",
```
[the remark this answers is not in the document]
gbowlin: If you don't want to do this, its probably better to revert back to a `python scripts/run_automation.py` and document the example rather than include the CLI in `setup.cfg`

src/seismometer/data/otel.py
```diff
@@ -0,0 +1,169 @@
...
+    from opentelemetry.sdk.metrics.export import ConsoleMetricExporter, PeriodicExportingMetricReader
+    from opentelemetry.sdk.resources import SERVICE_NAME, Resource
+
+    TELEMETRY_POSSIBLE = True
+except ImportError:
+    # No OTel.
+    TELEMETRY_POSSIBLE = False
+
+
+def read_otel_info(file_path: str) -> dict:
```
gbowlin, line 27: is this called anywhere anymore?

src/seismometer/api/plots.py
```diff
@@ -297,6 +327,29 @@ def _plot_leadtime_enc(
     good_groups = counts.loc[counts > censor_threshold].index
     summary_data = summary_data.loc[summary_data[cohort_col].isin(good_groups)]
 
+    log_all = otel.get_metric_config("Time Lead")["log_all"]
```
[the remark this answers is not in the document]
gbowlin: not relevant now

gbowlin (changes requested):
Looking good overall, but I ran into some issues with testing your updates. Can you include more details around configuration in the updated doucmentation? 

You may also want some default config in the example notebook to make it easier to get started.

docs/integration_guide/backend.rst
```diff
@@ -0,0 +1,132 @@
...
+.. code-block:: yaml
+
+  ports:
+    4317
+
+Start the instances with ``docker-compose up -d`` and then visit
+``localhost:4000`` to use Grafana. Log in with ``admin/admin``
+username/password to explore metrics.
+
+
```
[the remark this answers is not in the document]
adamhutchings: Done!

src/seismometer/core/autometrics.py
```diff
@@ -98,6 +106,46 @@ def get_function_from_export_name(self, fn_name: str) -> Callable:
...
+        plots and metrics they want to see). It also does not accommodate
+        cells being deleted, because this would require some more in-depth
+        access to the Jupyter frontend.
+        """
+        with open(self.automation_file_path, "w") as automation_file:
+            call_history = self._call_history
+            yaml.dump(call_history, automation_file)
+
+    def get_metric_config(self, metric_name: str) -> dict:
+        """_summary_
```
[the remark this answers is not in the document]
adamhutchings: Done.

docs/integration_guide/backend.rst
```diff
@@ -0,0 +1,132 @@
...
+Make sure your `config.yml`, in the `log:` section, has the following:
+.. code-block:: yaml
+
+  ports:
+    4317
+
+Start the instances with ``docker-compose up -d`` and then visit
+``localhost:4000`` to use Grafana. Log in with ``admin/admin``
+username/password to explore metrics.
```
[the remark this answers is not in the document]
adamhutchings: I gave a simple example.

src/seismometer/configuration/export_config.py
```diff
@@ -32,6 +35,11 @@ def __init__(self, raw_config: dict):
         else:
             self.otel_stdout = False
 
+        if "hostname" in log_config:
+            self.hostname = log_config["hostname"]
+        else:
+            self.hostname = "otel-collector"
```
gbowlin, line 41: rather than default, can you instead throw a ValueError?

src/seismometer/data/otel.py
```diff
@@ -98,7 +100,7 @@ def __init__(self, file_output_paths, export_ports, dump_to_stdout):
         for export_port in export_ports:
             otel_collector_reader = None
             try:
-                with socket.create_connection(("otel-collector", export_port), timeout=1):
+                with socket.create_connection((hostname, export_port), timeout=1):
```
gbowlin, line 103: I'd recommend breaking this out to enable better testing, so that you can mock the result and provide better error handling:


```python

def wait_for_socket(host: str, port: int, timeout: float = 1.0, max_retries: int = 10):
    for attempt in range(max_retries):
        try:
            with socket.create_connection((host, port), timeout=timeout):
                print(f"Connection to {host}:{port} succeeded.")
                return True
        except (socket.timeout, ConnectionRefusedError, OSError) as e:
            print(f"Attempt {attempt + 1} failed: {e}")
            time.sleep(1)  # Backoff before retrying
    raise TimeoutError(f"Could not connect to {host}:{port} after {max_retries} attempts.")
```

docs/integration_guide/metrics.rst
```diff
@@ -0,0 +1,90 @@
...
+
+A sequence of entries follows, each of which specifies a type of metric to be
+logged (accuracy, specificity, etc). There are four settings for each metric:
+
+- ``output_metrics``: whether metrics are to be output from this at all.
+- ``log_all``: in some displays, an entire curve is plotted while only a few
+  points are singled out (by threshold, for example). This options says whether
+  to log the entire curve or just the singled-out points.
+- ``granularity``: for plots which display in quantiles, specifies how many quantiles
+  to output in the metrics. (For instance, quartiles would be ``granularity: 4``.)
```
gbowlin, line 36: Reading this suggests a good rename `granularity` -> `quantiles`

`quantiles: 4` immediately suggests quartiles.

src/seismometer/data/otel.py
adamhutchings: Done.

src/seismometer/data/metric_apis.py
```diff
@@ -0,0 +1,260 @@
...
+
+    TELEMETRY_POSSIBLE = True
+except ImportError:
+    # No OTel.
+    TELEMETRY_POSSIBLE = False
+
+
+class OpenTelemetryRecorder:
+    def __new__(cls, *args, **kwargs):
+        if TELEMETRY_POSSIBLE:
```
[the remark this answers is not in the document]
adamhutchings: True! The other instance of TELEMETRY_POSSIBLE is harder to do this way because said imports are actually used outside of the __new__ scope, but this one has been fixed.

docs/integration_guide/metrics.rst
```diff
@@ -84,7 +84,10 @@ Such a file looks like a list of plot functions and their arguments, i.e.
         threshold: 0.1
 
 The ``options`` section specifies function arguments besides cohorts,
-and the cohorts themselves are specified in the ``cohorts`` section.
+and the cohorts themselves are specified in the ``cohorts`` section. The
+``extra_info`` section is yet unused but may in the future provide extra
+information about the context of a function call should we ever need
+to store such data.
```
gbowlin, line 90: if its unused, maybe we should just remove it?

adamhutchings: It has been removed.

src/seismometer/api/plots.py
```diff
@@ -297,6 +298,30 @@ def _plot_leadtime_enc(
     good_groups = counts.loc[counts > censor_threshold].index
     summary_data = summary_data.loc[summary_data[cohort_col].isin(good_groups)]
 
+    am = AutomationManager()
```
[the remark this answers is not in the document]
adamhutchings: I made it its own function there.

docs/integration_guide/metrics.rst
```diff
@@ -0,0 +1,88 @@
...
+===============================
+Individual Metric Configuration
+===============================
+
+The ``otel_info:`` section in the usage config file contains information about
+what metrics we want to export from our plotting functions.
+
+.. code-block:: yaml
+
+    otel_info:
```
gbowlin, line 21: Is there a way for us to make the default configuration to allow all metrics with log_all: false?

docs/integration_guide/metrics.rst
```diff
@@ -0,0 +1,88 @@
...
+      Accuracy:
+        output_metrics: true
+        log_all: false
+        quantiles: 4
+        measurement_type: Gauge
```
gbowlin, line 26: Are `quantiles` and `measurement_type` interlinked (can a Gauge support quantiles, or just histogram?)

adamhutchings: No. Theoretically, the measurement type just applies to how we aggregate metrics from multiple calls to metric population, whereas quantiles refers to the types of metrics we are collecting in fact.

gbowlin (changes requested):
This is really just documentation update. Otherwise its just a series of minor renames.

docs/integration_guide/metrics.rst
```diff
@@ -0,0 +1,88 @@
+================
+Metric Exporting
+================
+
+As OpenTelemetry has capabilities for exporting metrics from notebook runs,
+much information about metrics may be configured. There are two main places that
+metrics are configured:
+
+- The automation settings (likely in ``metric-automation/yml``)
```
gbowlin, line 9:
```suggestion
- The automation settings (likely in ``metric-automation.yml``)
```

docs/integration_guide/backend.rst
```diff
@@ -0,0 +1,201 @@
...
+  scrape_interval: 15s  # Frequency of metric scraping
+
+  scrape_configs:
+    - job_name: 'otel-collector'
+      static_configs:
+        - targets: ['otel-collector:9464']
+
+.. code-block:: yaml
+
+  # otel-collector-config.yml
```
gbowlin, line 95:
```suggestion
  # otel-collector-config.yaml
```

To match with the lines 63-65. You can reverse this if you like so all use `yml` as the extension of `yaml`

gbowlin: This was actually the error I was seeing earlier with configuration. I created a `otel-collector-config.yml`, but the docker compose example uses a different name `otel-collector-config.yaml`, causing the container to fail to load up, which caused export to fail as well. 

So long as its all consistent things work. 

[2025-08-01 18:48:06 UTC] DEBUG: Created cohorts: Age, Race, Gender, A1C, Taking Insulin, Taking Metformin, H1, H2, H3, H4, H5_source, H5
[2025-08-01 18:48:06 UTC] INFO: Connection to otelcollector:4317 succeeded.

docs/integration_guide/backend.rst
```diff
@@ -182,19 +182,17 @@ automatic export based on the saved settings using
...
 
 .. code-block:: python
 
   import seismometer as sm
-  # do not download any example dataset (like in example-notebooks) because this will
-  # overwrite your config for automation paths
   sm.run_startup(config_path='.') # load config
   sm.activate_exports()
-  sm.do_metric_exports()
+  sm.export_automated_metrics()
```
gbowlin, line 195: since `export_automated_metrics()` calls `activate_exports()` you can remove the line above now. 


```python
@export
def export_automated_metrics() -> None:
    """This function does automated metric exporting for
    everything specified in Seismogram.
    """
    from seismometer.data.otel import activate_exports

    activate_exports()
```

gbowlin (approved):
Looks good, thanks for the hard work!

PR #188, opened 2026-03-24
fix: relax censor_min_count lower bound from 10 to 0 (#182)

### LINKED ISSUE #182: Relax `censor_min_count` Constraint to Allow Thresholds Below 10 for Small/Synthetic Datasets
## Problem Summary
The `censor_min_count` configuration parameter is currently defined with a `ge=10` constraint, preventing users from setting a threshold below 10. This makes it impossible to generate visualizations on small or highly filtered datasets, even when privacy constraints do not apply (e.g., synthetic data, debugging, or unit tests).

## Impact
Users working with small cohorts, synthetic datasets, or temporary filtered subsets cannot produce visualizations because valid use cases require lower censoring thresholds. In particular, for certain visualization types — such as the `Likert plots` used to show categorical data — it may be perfectly reasonable to plot distributions even when counts are less than 10.

## Possible Solution
Relax the schema-level constraint so that users can explicitly set `censor_min_count` below 10 when appropriate. The default can remain 10, but the configuration should not forbid smaller values. Higher-level logic or warnings can still signal low-N situations without blocking visualization generation.

## Steps to Reproduce
1. Set `censor_min_count` to less than 10 in `usage_config` 
2. `sm.run_startup` fails.

## Suggested fix
Modify the Pydantic field definition for censor_min_count to allow values below 10. For example:
```python
censor_min_count: int = Field(10, ge=0)
```
Keep the default value at 10 so existing users and workflows remain unchanged but allow users to override the threshold for small-n visualizations (e.g. Likert plots or exploratory debugging).

d-koding:
# Overview
Closes #182

## Description of changes
Relaxes the `censor_min_count` lower bound from `ge=10` to `ge=0` in `DataUsage`.

The previous hard minimum of 10 blocked legitimate workflows where privacy constraints 
do not apply, such as synthetic data, debugging, or LLM feedback visualizations 
(e.g. Likert plots). The default value remains 10 so no existing behavior is changed.

- Changed `Field(10, ge=10)` to `Field(10, ge=0)` in `src/seismometer/configuration/model.py`
- Updated docstring with `.. versionchanged:: NEXT_VERSION_PLACEHOLDER` directive
- Added tests covering values below 10 and rejection of negative values
- Added changelog fragment `changelog/182.bugfix.rst`

## Author Checklist
- [x] Linting passes; run early with pre-commit hook.
- [x] For breaking API changes, added `.. versionchanged:: NEXT_VERSION_PLACEHOLDER` directive to affected docstrings.
- [x] Draft your news fragment in new `changelog/ISSUE.TYPE.rst` files; see changelog/README.md.


src/seismometer/configuration/model.py
```diff
@@ -496,7 +496,7 @@ class DataUsage(BaseModel):
     """A list of all metrics to load."""
     load_time_filters: list[FilterConfig] = []
     """A list of filters to apply at load time to reduce the working dataset."""
-    censor_min_count: int = Field(10, ge=10)
+    censor_min_count: int = Field(10, ge=0)
```
MahmoodEtedadi, line 499: Maybe we could add a validator to log a warning if `censor_min_count < 10`. Something like:
```python
    @field_validator("censor_min_count")
    def warn_below_recommended_floor(cls, censor_min_count: int) -> int:
        """Warn when censoring is configured below the recommended minimum."""
        if censor_min_count < 10:
            logger.warning(
                f"censor_min_count is set to {censor_min_count}, below the recommended minimum of 10. "
                "This weakens small-cell suppression, which guards against re-identifying individuals "
                "in small cohorts. Only use values below 10 with data that does not identify real "
                "individuals (e.g., synthetic data)."
            )
        return censor_min_count
```

changelog/182.feature.rst
```diff
@@ -0,0 +1 @@
+Relaxed ``censor_min_count`` lower bound from 10 to 0 to support small-n workflows such as synthetic data, debugging, and LLM feedback visualizations (e.g. Likert plots). Default remains 10. (`#182 <https://github.com/epic-open-source/seismometer/issues/182>`__)
```
MahmoodEtedadi, line 1: Should we rename this file to `182.feature.rst`? I am not sure `bugfix` is the correct classification as the hard stop of 10 was set to guard against including data that could identify individuals in small cohorts.

tests/configuration/test_model.py
```diff
@@ -313,6 +317,11 @@ def test_reduce_events_to_unique_names(self, caplog):
         data_usage.events[1].source == "event2"
         data_usage.events[1].display_name == "Event 2"
 
+    @pytest.mark.parametrize("value", [0, 1, 5, 9])
+    def test_censor_min_count_allows_below_10(self, value):
+        data_usage = undertest.DataUsage(censor_min_count=value)
+        assert data_usage.censor_min_count == value
+
```
MahmoodEtedadi, line 324: We could add tests to capture the warning logged. For instance,

```python
   @pytest.mark.parametrize("value", [0, 1, 9])
    def test_censor_min_count_below_floor_warns(self, value, caplog):
        with caplog.at_level(logging.WARNING, logger="seismometer"):
            undertest.DataUsage(censor_min_count=value)
        assert "below the recommended minimum" in caplog.text

    @pytest.mark.parametrize("value", [10, 20])
    def test_censor_min_count_at_or_above_floor_does_not_warn(self, value, caplog):
        with caplog.at_level(logging.WARNING, logger="seismometer"):
            undertest.DataUsage(censor_min_count=value)
        assert "below the recommended minimum" not in caplog.text
```

MahmoodEtedadi (commented):
Sorry for the late review. Just added a couple of minor considerations.

changelog/182.feature.rst
d-koding: Yes that makes more sense- it's a new capability, not a correction of unintended behavior, so 182.feature.rst is a more accurate classification.

src/seismometer/configuration/model.py
d-koding: If this validator fires on every model instantiation (e.g. once per config load, not once per row), I think this is a good change. But if censor_min_count is validated somewhere that gets called repeatedly, it'll spam the log- what do you think?

tests/configuration/test_model.py
d-koding: Yes this is a great idea, one thought "below the recommended minimum" in caplog.text is a reasonable substring check, but we could consider asserting on something that also captures the value, to make sure the interpolation actually worked:
in caplog.text
```suggestion
pythonassert f"censor_min_count={value} is below the recommended minimum" 
```

d-koding (commented):
In agreement with everything here, one small edit on the pytest could be effective but otherwise I'm happy to merge this.

MahmoodEtedadi (changes requested):
Agreed. Just to confirm, have you pushed your recent changes? I think we still have the old version here.

d-koding:
All changes were added in today- had to do it manually but all tests passed and it should be up to date with your suggestions

MahmoodEtedadi (approved):

